PR #2, opened 2015-01-30
fix Drive's xlsx conversion issue.

shibukawa:
When calling Drive.Files.Insert() with Excel 2007 (Open XML Format) and Convert(true) option, Google Drive ignores converting option.

Expected HTTP request (multipart) is the following:

```
--3c664c807714f0f19cbe60d94d7b6b69e7676da09d2f07b8bb4ce2df6c96

Content-Type: application/json

{"mimeType":"application/vnd.openxmlformats-officedocument.spreadsheetml.sheet","title":"test_xlsx"}

--3c664c807714f0f19cbe60d94d7b6b69e7676da09d2f07b8bb4ce2df6c96

Content-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet

 (xlsx body)
```

But googleapi.ConditionallyIncludeMedia() added auto detected MIME (application/zip) as a second part Content-Type. My fix adds as same MIME type as client code explicitly adds. Now Drive's Spreadsheet conversion works fine.

### DIFF
--- a/drive/v2/drive-gen.go
+++ b/drive/v2/drive-gen.go
@@ -3804,7 +3804,7 @@ func (c *FilesInsertCall) Do() (*File, error) {
 	urls += "?" + params.Encode()
 	if c.protocol_ != "resumable" {
 		var cancel func()
-		cancel, _ = googleapi.ConditionallyIncludeMedia(c.media_, &body, &ctype)
+		cancel, _ = googleapi.ConditionallyIncludeMediaWithMime(c.media_, c.file.MimeType, &body, &ctype)
 		if cancel != nil {
 			defer cancel()
 		}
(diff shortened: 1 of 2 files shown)


gmlewis:
We have a new Media implementation coming soon that will enable the content type to be explicitly specified, so I think we will not move forward with this patch... Feel free to continue using it locally until the new version is ready, of course.

gmlewis:
Additionally, we do not use pull requests. Github is only a mirror. See https://github.com/google/google-api-go-client/blob/master/CONTRIBUTING.md

shibukawa:
Thank you for your reply.

My goal of this Pull-Request is to notify the existence and the cause of the issue (because there is no issue page of this module). I look forward to use new Media code :)

gmlewis:
You can also currently use a ContentTyper that overrides the Content-Type:
https://github.com/shibukawa/google-api-go-client/blob/ce366bef7ecdbde760cb97f7c0ac4f039ccdf864/googleapi/googleapi.go#L30
You can see an example of its usage in the Google Cloud Storage veneer library:
https://github.com/GoogleCloudPlatform/gcloud-golang/blob/master/storage/types.go#L307

PR #83, opened 2015-05-19
Generate all bool fields in structs as pointers to allow sending false

ppg:
The currently generated code creates structs with value type fields and the `omitempty` directive in the JSON tag. There is discussion around this as it relates to fields that have a default of true https://github.com/google/google-api-go-client/issues/54, namely that they can never be changed. However, the issue is really more broad, since `omitempty` is omitting that field if the value is the field type's zero value. That means for any bool field you can set it to `true`, but you can never set it to `false` through the API, since the serialization will always omit that. This is really an issue for all fields, i.e. you could never set a string to `""`, but its especially poignant for bools, as you're prevented from half of the possible values to set. The specific example we currently have is wanting to unsuspend a user, which the API supports with `suspend=false` passed along, but there's no way for the Go API client to generate that body.

Long term, this generated code probably needs to behave more like gRPC where all fields are pointers so they can be zeroed or nillified, and some helper methods are provided like GetSuspended, that will default to zero value when nil.


bradfitz:
Like Go, we don't use Github for code review. We use Gerrit instead. We only use Github as a mirror and for its issue tracker.

See https://golang.org/doc/contribute.html except our upstream is https://code.googlesource.com/google-api-go-client/ (so replace "go.googlesource" with "code.googlesource" (or "go-review" with "code-review") and "google-api-go-client" instead of "go").

We should probably have more specific docs for how to contribute to this project but for a long time it was only me, and then recently @gmlewis and @mcgreevy.

Glenn, you want to add a CONTRIBUTING.md to this repo with tips?

In the meantime, closing this PR because we don't use PRs. We have to review and submit this upstream. Could you move it there?

/cc @gmlewis @mcgreevy

PR #100, opened 2015-08-28
Update README.md

peterkshultz:
Indentation needed so as to maintain style consistency with all other links.


bradfitz:
We don't use pull requests. See https://github.com/google/google-api-go-client/blob/master/CONTRIBUTING.md

peterkshultz:
Apologies!

gmlewis:
Thanks for the report.  I've cleaned up the README.md here:
https://code-review.googlesource.com/#/c/3352/
One of the problems was that it had tab characters in many places, which I've removed.

PR #121, opened 2016-01-14
translate: fixed detection and translation calls

zippoxer:
Those didn't work before due to a little JSON mistake.

It's working for me now, so I thought I'd share it.


gmlewis:
This repo does not take pull requests.  In fact, the file you modified is auto-generated so any changes would need to be made to the generator in such a manner that none of the other APIs are affected.

Please see https://github.com/google/google-api-go-client/blob/master/CONTRIBUTING.md for more information

PR #235, opened 2017-10-20
Use WithoutDataWrapper instead of WithDataWrapper

akm:
Because POST https://translation.googleapis.com/language/translate/v2 doesn't need `data`
https://cloud.google.com/translate/docs/reference/translate


akm:
I signed it!

mattwhisenhunt:
Thanks for your contribution!
This repo does not take pull requests. Please see https://github.com/google/google-api-go-client/blob/master/CONTRIBUTING.md for more information.

mattwhisenhunt:
@akm It looks like this response is wrapped in a data field (see Response body section) so I don't think this change is correct at any rate.

akm:
@mattwhisenhunt Okay, I'll close this. Thanks.

PR #242, opened 2017-11-17
Update docstring to reflect what the function actually does

avishalom:


bradfitz:
Note that this PR can't be accepted here, as this is a mirror repo. The upstream lives in Gerrit.
 
/cc @jba

jba:
Thanks for your contribution!

I made this change in https://code-review.googlesource.com/#/c/google-api-go-client/+/20531.

PR #246, opened 2017-12-29
changes ml-gen.go

carlosrgomes:
Changes to the client go of the ML learning machine to function correctly.
I was using api rest in appengine with golang and it does not work if you do not have these changes.


bradfitz:
Sorry, this is an auto-generated file. We don't accept changes to it. Also, we don't accept changes via GitHub currently, as this is just a mirror. You would need to change the generator.

I recommend you file a bug instead and somebody on the team can figure out what's going on with App Engine.

PR #248, opened 2018-01-17
Update getting started file to use new creds page

jadekler:
Note: the old url redirects to the new, but I think it might be nice just to directly have the new


jadekler:
Forgetting that this uses gerrit

Issue #241, opened 2017-11-09
The ML engine library does not work.


Mistobaan:
I have to manually patch the library to make it work:

```diff
diff --git a/ml/v1/ml.go b/ml/v1/ml.go
index 30573fa..498edb1 100644
--- a/ml/v1/ml.go
+++ b/ml/v1/ml.go
@@ -14,15 +14,16 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	context "golang.org/x/net/context"
-	ctxhttp "golang.org/x/net/context/ctxhttp"
-	gensupport "google.golang.org/api/gensupport"
-	googleapi "google.golang.org/api/googleapi"
 	"io"
 	"net/http"
 	"net/url"
 	"strconv"
 	"strings"
+
+	context "golang.org/x/net/context"
+	ctxhttp "golang.org/x/net/context/ctxhttp"
+	gensupport "google.golang.org/api/gensupport"
+	googleapi "google.golang.org/api/googleapi"
 )
 
 // Always reference these packages, just in case the auto-generated code
@@ -2731,15 +2732,21 @@ func (c *ProjectsPredictCall) doRequest(alt string) (*http.Response, error) {
 	}
 	reqHeaders.Set("User-Agent", c.s.userAgent())
 	var body io.Reader = nil
-	body, err := googleapi.WithoutDataWrapper.JSONReader(c.googlecloudmlv1__predictrequest)
-	if err != nil {
-		return nil, err
-	}
+	// body, err := googleapi.WithoutDataWrapper.JSONReader(c.googlecloudmlv1__predictrequest)
+	// if err != nil {
+	// 	return nil, err
+	// }
+	body = strings.NewReader(c.googlecloudmlv1__predictrequest.HttpBody.Data)
+
 	reqHeaders.Set("Content-Type", "application/json")
 	c.urlParams_.Set("alt", alt)
 	urls := googleapi.ResolveRelative(c.s.BasePath, "v1/{+name}:predict")
 	urls += "?" + c.urlParams_.Encode()
-	req, _ := http.NewRequest("POST", urls, body)
+	req, err := http.NewRequest("POST", urls, body)
+	if err != nil {
+		return nil, err
+	}
+
 	req.Header = reqHeaders
 	googleapi.Expand(req.URL, map[string]string{
 		"name": c.name,
[...]

PR #285, opened 2018-07-31
Fixed - #241 The ML engine library does not work. 

shinofara:
## Issue

https://github.com/google/google-api-go-client/issues/241

## Change point

- Add Error Handring
- Changing parsing of responses


broady:
Hi, we don't use GitHub pull requests.

See CONTRIBUTING for instructions:
https://github.com/google/google-api-go-client/blob/master/CONTRIBUTING.md

(You already sent this patch over there, but I'm replying here in case other people see this.)

PR #286, opened 2018-08-01
add filename to mediaitem

skpenpen:


skpenpen:
I signed it!

jba:
This is generated code. We have to update the discovery doc (the json file). Do you have a recent one? It doesn't seem to be in the list of public docs.

Also, we use Gerritt for PRs, so unless you want to learn that, just point me to the discovery doc and I'll make the change.

skpenpen:
Do you mean https://github.com/skpenpen/google-api-go-client/blob/master/photoslibrary/v1/photoslibrary-api.json ?

jba:
I'm sorry, we can't accept this PR. You should file a feature request with the Photos Library API directly at  https://developers.google.com/photos/library/support/how-to-get-help.

PR #311, opened 2018-11-23
Fix README Travis Badge

grant:
Badge should show:
image


jadekler:
We're actually on the cusp of removing travis CI, so we'll pass on this. Thanks for the PR though Grant! :)

PR #314, opened 2018-11-28
Remove omitempty for those required fields

businiaowyf:
https://developers.google.com/youtube/v3/live/docs/liveBroadcasts/update
As the specification in youtube data api, when updating broadcast, some fields are required, we can't omit them.


businiaowyf:
I signed it!

jadekler:
Unfortunately we cannot make manual changes to these generated files: they would get blown away the next time we run the generator. The backend team has to publish the changes for the generator to pick them up. However, our generator has not run in a little bit, so it's possible they've published them and we just haven't updated our libraries. We're investigating why the generator is down.

Since we don't make manual changes to generator code, I'm going to close this PR for now. Give a shout if you want to discuss this further though!

PR #356, opened 2019-04-17
Improve diskTypes URL documentation in several places

kinghrothgar:
In several comments, `project/zones/zone/diskTypes/` is used as an example for diskType URL which is not valid. This led to confusion that took me about an hour to sort out. Instead, `projects/project/zones/zone/diskTypes/` should be used which IS valid.

### DIFF
--- a/compute/v0.alpha/compute-api.json
+++ b/compute/v0.alpha/compute-api.json
@@ -31280,7 +31280,7 @@
           "type": "string"
         },
         "type": {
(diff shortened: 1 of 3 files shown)


kinghrothgar:
To make it even more confusing, this invalid example can be found everywhere https://cloud.google.com/compute/docs/reference/rest/v1/disks/insert

kinghrothgar:
Just realized this is auto generated, so I'll just move this to an Issue

PR #377, opened 2019-07-07
Initial WhiteSource configuration file

Dollar09:
Future/cz”&


bradfitz:
It looks like this commit was also a mistake? If not, send it again with an explanation.

PR #385, opened 2019-08-01
Fix relative path for ProjectsEventsReportCall

gm42:
There is at least one occurrence of invalid URI when attempting to report an error, which results in a 404.

With the proposed fix the issue is solved; this is however, likely incomplete as other URIs might have the same issue and generally the generator of this code should be fixed instead.

Documentation confirms the correct URL: https://cloud.google.com/error-reporting/reference/rest/v1beta1/projects.events/report


gm42:
@googlebot I signed it!

jadekler:
Sorry, we do not accept PRs on github (see our CONTRIBUTING.md). Also, that file is auto-generated - changing it would result in your change being blown away during next regen.

Would you mind filing an issue about your experience, with a small reproduction?

gm42:
>  that file is auto-generated - changing it would result in your change being blown away during next regen.

Yes, I mentioned in the description that it should be solved in the generator possibly.

@jadekler sorry, I did not check `CONTRIBUTING.md`. I will create an issue.

PR #429, opened 2019-11-25
Ensure that any bodies created in `GetBody` are cleaned up

kevinconaway:
This PR fixes issues brought up in:

https://github.com/googleapis/google-cloud-go/issues/1651
https://github.com/googleapis/google-cloud-go/issues/1380

As best I understand it, the goroutines will leak when `GetBody` is called and the resulting body is not closed.

This PR modifies `UploadRequest` to add any request bodies created by `GetBody` to a list to be cleaned up by the `cleanup` function that is returned.  Previously, only the main request would be cleaned up but not any of those returned by `GetBody`.

I've verified that this fixes the issue in our environment.

### DIFF
--- a/internal/gensupport/media.go
+++ b/internal/gensupport/media.go
@@ -290,6 +290,9 @@ func (mi *MediaInfo) UploadRequest(reqHeaders http.Header, body io.Reader) (newB
 		fb := readerFunc(body)
 		fm := readerFunc(media)
 		combined, ctype := CombineBodyMedia(body, "application/json", media, mi.mType)
+		toCleanup := []io.Closer{
+			combined,
+		}
 		if fb != nil && fm != nil {
 			getBody = func() (io.ReadCloser, error) {
 				rb := ioutil.NopCloser(fb())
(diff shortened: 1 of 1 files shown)


kevinconaway:
@googlebot I signed it!

jadekler:
Hello! Thank you for investigating this.

@cjcotter @frankyn - the biggest client that uses this is storage AFAIK. Could one of you investigate this with fairly high priority? This may fix some of the other subtle issues we've been seeing.

@kevinconaway Unfortunately this github repo is just a mirror of another git repo that uses gerrit for code review. Would you mind closing this PR and submitting a change request to gerrit? Our CONTRIBUTING.md has extensive guidance, and I'm happy to help too. If you prefer not, I can submit this for you, but unfortunately then you wouldn't get credit which is not as cool.

kevinconaway:
Thanks, I'll do so.

kevinconaway:
Created a gerrit CL here https://code-review.googlesource.com/c/google-api-go-client/+/48910

PR #442, opened 2020-01-13
option: remove experimental API WithGRPCConnectionPool

### LINKED ISSUE #441: grpc.RoundRobin and grpc.WithBalancer no longer exist
We are using this library and seeing these errors:

../google.golang.org/api/option/option.go:153:14: undefined: grpc.RoundRobin
../google.golang.org/api/option/option.go:154:42: undefined: grpc.WithBalancer

Sounds like gRPC removed these legacy APIs: https://github.com/grpc/grpc-go/commit/336cf8d76145dc5ebd517fd9c19e14c6822450b3

dfawley:
The experimental APIs relied upon by this option were deleted in https://github.com/grpc/grpc-go/pull/3301

Fixes #441


dfawley:
@jadekler - can you take a look?

jadekler:
Ooo, yikes, this one is bad. I'm OOO and also no longer on client libraries - paging @tbpg, @codyoss, @broady .

broady:
-2 on this CL. I'll work on this. Thanks for raising.

codyoss:
Note we do make use of the option in google-cloud-go in a couple of spots as well.

dfawley:
This is an immediate fix to all users that do not depend upon the experimental parts of this package.  I would recommend submitting it to unbreak people while you work on a full fix.

Note that any package which contains a dependency on any experimental code should be considered experimental in its entirety.  So I would recommend re-adding the experimental functionality in a separate package.

broady:
as discussed, fix is in progress

PR #490, opened 2020-05-20
User agent prefix

kenegozi:
Per the RFC at https://tools.ietf.org/html/rfc2616#section-14.43, "By convention, the product tokens are listed in order of their significance for identifying the application."

Support specifying a prefix to the user agent, so that a user could specify a more significant identifier for their requests.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -43,6 +43,7 @@ Glenn Lewis <gmlewis@google.com>
 Ivan Krasin <krasin@golang.org>
 Jason Hall <jasonhall@google.com>
 Johan Euphrosine <proppy@google.com>
+Ken Egozi <kenegozi@google.com>
 Kostik Shtoyk <kostik@google.com>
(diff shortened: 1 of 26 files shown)


broady:
It sounds like we should simply swap the order to match the spec. Also, some more ideas here: #494

kenegozi:
imo when *setting* a `UserAgent` field the expected behavior is to *replace* the user agent, not to append or prepend. Changing it in any way would be a breaking change though which is why I proposed a new field explicitly called `UserAgentPrefix`.

Interestingly enough, `WithUserAgent()` in `NewService()` would *replace* the agent, however it would not work with a custom http client provided via `WithHttpClient`.

A reasonable way forward (instead of the one proposed here) would be to mark `UserAgent` as obsolete, and have `WithUserAgent` work with a provided http client. I am happy to work on that change if preferred.

blackshipsail:
@gopherbot

Issue #484, opened 2020-05-18
idtoken: should validate expiry and possibly issuer


broady:
Currently the idtoken package verifies audience and that the JWT is valid.

But it doesn't validate that `exp` is in the future.

It's possible we should check the issuer, too - it should be `https://accounts.google.com`

Updates #160.

codyoss:
In the case of IAP, the issuer is `https://cloud.google.com/iap`.

codyoss:
Also for other tokens it look like it can be one of two values...: `https://accounts.google.com`, `accounts.google.com`

https://developers.google.com/identity/sign-in/web/backend-auth

codyoss:
Also also, for ESP we say that the issuer can be a hostname or an email.[1](https://cloud.google.com/endpoints/docs/openapi/openapi-extensions#x-google-issuer) I am not sure we need to support this flow with this API though.

broady:
Hmmm, let's not check the issuer. The fact that it's signed by Google and unexpired I think is "enough" validation.

codyoss:
Awesome, as expiry has been added I will close this issue.

PR #492, opened 2020-05-20
idtoken: validate if token is expired

codyoss:
Updates: #484

### DIFF
--- a/idtoken/validate_test.go
+++ b/idtoken/validate_test.go
@@ -24,16 +24,14 @@ import (
 )
 
 const (
-	keyID        = "1234"
-	testAudience = "test-audience"
+	keyID              = "1234"
+	testAudience       = "test-audience"
+	expiry       int64 = 233431200
 )
 
 var (
-	expiry int64 = 233431200
-	// one second before the token expires
-	beforeExp = func() time.Time { return time.Unix(233431199, 0) }
-	// one second after the token expires
-	afterExp = func() time.Time { return time.Unix(233431201, 0) }
+	beforeExp = func() time.Time { return time.Unix(expiry-1, 0) }
+	afterExp  = func() time.Time { return time.Unix(expiry+1, 0) }
 )
 
 func TestValidateRS256(t *testing.T) {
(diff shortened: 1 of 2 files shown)


idtoken/validate_test.go
```diff
@@ -27,18 +28,27 @@ const (
...
+		nowFunc func() time.Time
 		wantErr bool
 	}{
-		{name: "works", keyID: keyID, n: pk.N, e: pk.E, wantErr: false},
-		{name: "no matching key", keyID: "5678", n: pk.N, e: pk.E, wantErr: true},
-		{name: "sig does not match", keyID: keyID, n: new(big.Int).SetBytes([]byte("42")), e: 42, wantErr: true},
+		{name: "works", keyID: keyID, n: pk.N, e: pk.E, nowFunc: beforeExp, wantErr: false},
+		{name: "no matching key", keyID: "5678", n: pk.N, e: pk.E, nowFunc: beforeExp, wantErr: true},
+		{name: "sig does not match", keyID: keyID, n: new(big.Int).SetBytes([]byte("42")), e: 42, nowFunc: beforeExp, wantErr: true},
+		{name: "token expired", keyID: keyID, n: pk.N, e: pk.E, nowFunc: afterExp, wantErr: true},
```
tbpg, line 52: Optional: might be easier to read these if they were on separate lines.

idtoken/validate_test.go
```diff
@@ -64,7 +74,7 @@ func TestValidateRS256(t *testing.T) {
 					}
 				}),
 			}
-
+			now = tt.nowFunc
```
tbpg, line 78: Reset after the test? Don't want to accidentally mess up another test.

idtoken/validate_test.go
```diff
@@ -27,18 +28,27 @@ const (
 	testAudience = "test-audience"
 )
 
+var (
+	// one second before the token expires
+	beforeExp = func() time.Time { return time.Unix(233431199, 0) }
```
broady, line 33: maybe

```
var (
	expiry    = ...
	beforeExp = func() time.Time { return exp + time.Second }
	afterExp  = func() time.Time { return exp + time.Second }
)
```

codyoss: Done

idtoken/validate_test.go
```diff
@@ -27,18 +28,28 @@ const (
 	testAudience = "test-audience"
 )
 
+var (
+	expiry int64 = 233431200
+	// one second before the token expires
+	beforeExp = func() time.Time { return time.Unix(233431199, 0) }
+	// one second after the token expires
+	afterExp = func() time.Time { return time.Unix(233431201, 0) }
```
tbpg, line 36: Make this a const, then reference it in `beforeExp` and `afterExp`?

idtoken/validate_test.go
codyoss: Good point. Done.

idtoken/validate_test.go
codyoss: I like it, done.

idtoken/validate_test.go
codyoss: Yes, done.

PR #496, opened 2020-05-21
transport/http: update default transport settings

tritone:
Updates the default HTTP transport to use a larger value for
MaxIdleConnsPerHost. This improves performance under high load
for the GCS client.

Copied from gerrit CL https://code-review.googlesource.com/c/google-api-go-client/+/55470


broady:
This actually has another effect: the transport (`http.DefaultTransport`) would now no longer be shared between clients.

I don't know if that matters, but if it does, we would want the tweaked transport stored as a package-level variable rather than being copied every time a new client is created.

Otherwise, this looks good. I would tweak the build tags a bit:

```go
//====dial.go===

// defaultBaseTransport returns the base HTTP transport.
// On App Engine, this is urlfetch.Transport.
// Otherwise, use a default transport, taking most defaults from
// http.DefaultTransport.
// If TLSCertificate is available, set TLSClientConfig as well.
func defaultBaseTransport(ctx context.Context, clientCertSource cert.Source) http.RoundTripper {
	if appengineUrlfetchHook != nil {
		return appengineUrlfetchHook(ctx)
	}

	trans := clonedTransport(http.DefaultTransport)
	if trans == nil {
		trans = fallbackBaseTransport()
	}
	trans.MaxIdleConnsPerHost = 100

	if clientCertSource != nil {
		trans.TLSClientConfig = &tls.Config{
			GetClientCertificate: clientCertSource,
		}
	}

	return trans
}

// fallbackBaseTransport is used in <go1.13 as well as in the rare case if http.DefaultTransport has been reassigned something that's not a *http.Transport.
func fallbackBaseTransport() *http.Transport {
	return &http.Transport{
		Proxy: http.ProxyFromEnvironment,
		DialContext: (&net.Dialer{
			Timeout:   30 * time.Second,
			KeepAlive: 30 * time.Second,
			DualStack: true,
		}).DialContext,
		MaxIdleConns:          100,
		MaxIdleConnsPerHost:   100,
		IdleConnTimeout:       90 * time.Second,
		TLSHandshakeTimeout:   10 * time.Second,
		ExpectContinueTimeout: 1 * time.Second,
	}
}


//====default_transport_go113.go===

// clonedTransport returns the given RoundTripper as a cloned *http.Transport.
// It returns nil if the RoundTripper can't be cloned or coerced to *http.Transport.
func clonedTransport(rt http.RoundTripper) *http.Transport {
	t, ok := rt.(*http.Transport)
	if !ok {
		return nil
	}
	return t.Clone()
}


//====default_transport_not_go113.go===

func clonedTransport(rt http.RoundTripper) *http.Transport { return nil }
```

tritone:
Thanks for your comments, @broady! I fixed the cloning and build tag factoring per your suggestion.

About the fact that the default transport is no longer shared between clients-- do you think I should address that in this PR? Let me know what you think.

transport/http/dial.go
```diff
@@ -162,23 +164,46 @@ var appengineUrlfetchHook func(context.Context) http.RoundTripper
...
 func defaultBaseTransport(ctx context.Context, clientCertSource cert.Source) http.RoundTripper {
 	if appengineUrlfetchHook != nil {
 		return appengineUrlfetchHook(ctx)
 	}
 
+	// Copy http.DefaultTransport except for MaxIdleConnsPerHost setting,
+	// which is increased due to reported performance issues under load in the GCS
+	// client. Transport.Clone is only available in Go 1.13 and up.
+	trans := clonedTransport(http.DefaultTransport)
+	trans.MaxIdleConnsPerHost = 100
```
codyoss, line 179: This is nil in <go1.13.

transport/http/dial.go
```diff
@@ -162,23 +164,46 @@ var appengineUrlfetchHook func(context.Context) http.RoundTripper
...
 	}
 
-	return http.DefaultTransport
+	return trans
+}
+
+// fallbackBaseTransport is used in <go1.13 as well as in the rare case if
+// http.DefaultTransport has been reassigned something that's not a
+// *http.Transport.
+func fallbackBaseTransport() *http.Transport {
```
codyoss, line 193: This does not appear to be used, but I think it should

tritone: Whoops, thanks for catching this. Fixing now.

transport/http/dial.go
tritone: Fixed this, thanks!

transport/http/dial.go
```diff
@@ -162,23 +164,49 @@ var appengineUrlfetchHook func(context.Context) http.RoundTripper
...
+		trans.TLSClientConfig = &tls.Config{
+			GetClientCertificate: clientCertSource,
 		}
 	}
 
-	return http.DefaultTransport
+	return trans
+}
+
+// fallbackBaseTransport is used in <go1.13 as well as in the rare case if
```
broady, line 193:
```suggestion
var fallback struct {
	*http.Transport
	sync.Once
}

func fallbackBaseTransport() *http.Transport {
	fallback.Do(func() {
		fallback.Transport = &http.Transport{
			Proxy: http.ProxyFromEnvironment,
			DialContext: (&net.Dialer{
				Timeout:   30 * time.Second,
				KeepAlive: 30 * time.Second,
				DualStack: true,
			}).DialContext,
			MaxIdleConns:          100,
			MaxIdleConnsPerHost:   100,
			IdleConnTimeout:       90 * time.Second,
			TLSHandshakeTimeout:   10 * time.Second,
			ExpectContinueTimeout: 1 * time.Second,
		}
	})
	return fallback.Transport
}
```

broady:
> About the fact that the default transport is no longer shared between clients-- do you think I should address that in this PR? Let me know what you think.

Yeah, probably worth doing. Sent some code suggestion.

transport/http/dial.go
tritone: Looks good, thanks!

tritone: Took your suggestion on this.

PR #499, opened 2020-05-21
internal/kokoro/synth: fix compilation errors

broady:
#489 had code that didn't compile (whoops), but strangely didn't get picked up by tests.


codyoss:
This should be caught by our CI now.

codyoss:
Merging to fix CI.

PR #502, opened 2020-05-21
internal/kokoro: print out commit hash for HEAD

broady:


internal/kokoro/test.sh
```diff
@@ -15,6 +15,8 @@ set -x
 # cd to project dir on Kokoro instance
 cd git/google_api
 
+git rev-parse HEAD
```
tbpg, line 18: This should be printed by Kokoro. For example

```
Note: checking out '0621fd1d165ef16d7fdb53720d232c894dde22a4'.
```

From https://source.cloud.google.com/results/invocations/0f3145e4-2b9f-447f-bd03-4c21011daf56/log.

But, good with me if you want to print it here.

broady: Oh, missed that.

broady:
Okay... there's the weird thing. The build log associated with this PR says:
```
Note: checking out '0621fd1d165ef16d7fdb53720d232c894dde22a4'.
```

which is HEAD@master right now. I guess the presubmit config is messed up?

codyoss:
Now that we know what was going on, I think we can close this?

PR #503, opened 2020-05-21
internal/kokoro: fix CI pathing

codyoss:


codyoss:
Using admin priv to merge this. Build is failing but should be fixed by #499

PR #510, opened 2020-05-27
internal/kokoro: fix typo in script

codyoss:
Fixes: #509

### DIFF
--- a/internal/kokoro/synth/synth.sh
+++ b/internal/kokoro/synth/synth.sh
@@ -24,7 +24,7 @@ pushd google-api-go-generator
 make all
 popd
 
-if [ -z "$(git status --porcelain)"]; then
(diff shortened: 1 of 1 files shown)


internal/kokoro/synth/synth.sh
```diff
@@ -24,7 +24,7 @@ pushd google-api-go-generator
 make all
 popd
 
-if [[ -z "$(git status --porcelain)"]]; then
+if [ -z "$(git status --porcelain)"]; then
```
tbpg, line 27: I think we can keep the two brackets, but we need a space between " and ].
```suggestion
if [[ -z "$(git status --porcelain)" ]]; then
```

codyoss: Done

PR #513, opened 2020-05-29
internal: refactor replacement pkgs and remove monitoring v1

codyoss:
Refactored this bit of code so it is a little more flexible. This
now allows more granular control of which versions should be
deprecated.

Monitoring v1 is still receives updates today and has some apis
that are not captured by v3. Therefore, I am removing the
deprecation notice. This was brought to my attention in
https://github.com/googleapis/google-cloud-go/issues/2213.


codyoss:
In general I wonder if we should re-consider this approach of deprecating clients in favor of gocloud. If we really do want to do this, I think we need a more wholistic plan for doing this for more(all) apis.

google-api-go-generator/replacements.go
```diff
@@ -4,31 +4,121 @@
...
-	"pubsub":            "cloud.google.com/go/pubsub",
-	"redis":             "cloud.google.com/go/redis/apiv1",
-	"spanner":           "cloud.google.com/go/spanner",
-	"speech":            "cloud.google.com/go/speech/apiv1",
-	"texttospeech":      "cloud.google.com/go/texttospeech/apiv1",
-	"translate":         "cloud.google.com/go/translate",
-	"videointelligence": "cloud.google.com/go/videointelligence/apiv1",
-	"vision":            "cloud.google.com/go/vision/apiv1",
-	"storage":           "cloud.google.com/go/storage",
+var replacementPackage deprecatedPkgs = map[string]*apiReplacement{
```
tbpg, line 29: Why pointers? Should we use "comma ok" instead?

google-api-go-generator/replacements.go
```diff
@@ -4,31 +4,121 @@
...
+	"bigquery": {
+		replacementPkg: "cloud.google.com/go/bigquery",
+		versions:       []string{"*"},
```
tbpg, line 32: I wonder if the default (no versions listed) should mean all?

google-api-go-generator/replacements.go
codyoss: No reason for pointer I suppose, that works for me.

google-api-go-generator/replacements.go
codyoss: I like that idea, will do.

google-api-go-generator/replacements.go
```diff
@@ -4,31 +4,104 @@
...
+	},
+	"language": {
+		replacementPkg: "cloud.google.com/go/language/apiv1",
+	},
+	"logging": {
+		replacementPkg: "cloud.google.com/go/logging",
+	},
+	"monitoring": {
+		replacementPkg: "cloud.google.com/go/monitoring/apiv3",
+		versions:       []string{"v3"},
```
broady, line 65: if all of this was to do different version selection could i suggest using this syntax:

```
"logging": ...., // matches any logging version
"monitoring:v3": ... // matches only v3 of monitoring

This makes it a bit closer to other bits of the generator for special behavior

If you continue with the struct, I suggest putting most of these entires on a single line, like this:
https://play.golang.org/p/kqkLL3Mb1GI

File: google-api-go-generator/replacements_test.go [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+	tests := []struct {
+		name      string
+		dp        deprecatedPkgs
+		inName    string
+		inVersion string
+		want      string
+	}{
+		{
+			name:      "replacement found",
+			dp:        map[string]string{"foo:v1": "example.com/foo"},
```
tbpg, line 19:
```suggestion
			dp:        deprecatedPkgs{"foo:v1": "example.com/foo"},
```

?

codyoss: Done.

PR #515, opened 2020-05-29
all: autogenerated update (2020-05-29)

codyoss:
Ran `make all` and fixed test.


codyoss:
Since are automation needs to be changed out, this required a manual fix, and clients have not been generated for some time. I kicked it off manually.

Issue #449, opened 2020-02-10
internal/gensupport: resumable upload doesn't retry network errors or unexpected EOF when wrapped in url.Error


rogerhub:
I use the [google-cloud-go GCS client](https://github.com/googleapis/google-cloud-go/tree/master/storage), and I noticed that object uploads were failing sometimes with error messages like:

```
Post https://www.googleapis.com/upload/storage/v1/b/...: write tcp [2001:db8::]:...->[2001:db8::]:443: write: connection reset by peer
```

The GCS library delegates retry behavior to this package (google-api-go-client) -- see googleapis/google-cloud-go#1507 and [this test case](https://github.com/googleapis/google-cloud-go/blob/master/storage/retry_test.go).

This package defines retryable errors as:

```go
	var shouldRetry = func(status int, err error) bool {
		if 500 <= status && status <= 599 {
			return true
		}
		if status == statusTooManyRequests {
			return true
		}
		if err == io.ErrUnexpectedEOF {
			return true
		}
		if err, ok := err.(interface{ Temporary() bool }); ok {
			return err.Temporary()
		}
		return false
	}
```

I did a bit of testing, and I noticed that these errors are considered non-retryable:

1) A `*url.Error`, where the embedded error (i.e. `err.Unwrap()`) is io.ErrUnexpectedEOF.
2) A `*url.Error`, where the embedded error (i.e. `err.Unwrap()`) is a `*net.OpError` containing "write: connection reset by peer".

Some thinking here:

(1) is probably unintentional, since shouldRetry explicitly mentions that io.ErrUnexpectedEOF should be retried.

(2) seems retryable to me (although this is debatable). The "net" package says that connection reset errors are not Temporary() -- unless the operation is socket accept. However, that just means that the socket should be discarded. In the context of the overall resumable upload operation, this error should be considered a temporary network error and should be retried.

Anyway, the presence of these two errors make it so that uploads are unexpectedly fragile.

Here's my crappy attempt at a fix: (this suffices for my own use case)

```patch
diff --git a/internal/gensupport/resumable.go b/internal/gensupport/resumable.go
index e67ccd9a6..aa63fc7fe 100644
--- a/internal/gensupport/resumable.go
+++ b/internal/gensupport/resumable.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"sync"
 	"time"
[...]

PR #529, opened 2020-06-12
gensupport: expand retryable errors

tritone:
This commit improves retry handling by allowing retries
for the following:
* Wrapped errors that are retryable (Go 1.13+ only)
* 429 errors
* Transient network errors

Following retry guidance from
https://cloud.google.com/storage/docs/exponential-backoff

Note that this conflicts with #528, but it's an easy conflict
for me to fix once one of them is merged.

Fixes googleapis#449


internal/gensupport/resumable.go
```diff
@@ -36,6 +37,11 @@ const (
 	// should be retried.
 	// https://cloud.google.com/storage/docs/json_api/v1/status-codes#standardcodes
 	statusTooManyRequests = 429
+
+	// statusRequestTimeout is returned by the storage API if the request times
+	// out on the server side. The request should be retried.
+	// https://cloud.google.com/storage/docs/json_api/v1/status-codes#standardcodes
```
codyoss, line 43: For storage this is documented safe to retry, but I do wonder if it is in generate. I am thinking of a case where your request gets to the backend, is fully processed, but the response is never received back on the client side. If an API is not idempotent the second request may not yield the expected result. Thoughts?

tritone: Good point-- in any case this error has not be a major focus of complaints so far (for writes in the go libraries anyway). I could drop this one for now if you think it's better.

codyoss: I think we would want to dive into this more before adding holistically. We should probably only generically retry things we have documented in one of our aips. This is the only one I found with a quick search that mentions retry codes: https://google.aip.dev/194

tritone: Makes sense-- I removed the retry on 408.

internal/gensupport/resumable.go
```diff
@@ -239,3 +225,31 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
...
+	// Transient network errors should be retried.
+	if se, ok := err.(syscall.Errno); ok {
+		return se == syscall.ECONNRESET || se == syscall.ECONNREFUSED
+	}
+	// If Go 1.13 error unwrapping is available, use this to examine wrapped
+	// errors.
+	if err, ok := err.(interface{ Unwrap() error }); ok {
+		return shouldRetry(status, err.Unwrap())
+	}
+	if err, ok := err.(interface{ Temporary() bool }); ok {
```
codyoss, line 251: Should temporary be checked first? I think working off of the current error before diving deeper may be desirable.

tritone: Good point-- reworked this so that if the current error is Temporary, we return from there rather than unwrapping

tritone:
Got a +1 from @rogerhub as well (reporter of #449).

internal/gensupport/resumable.go
```diff
@@ -239,3 +225,33 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
...
+	}
+	if status == statusTooManyRequests {
+		return true
+	}
+	if err == io.ErrUnexpectedEOF {
+		return true
+	}
+	// Transient network errors should be retried.
+	if se, ok := err.(syscall.Errno); ok {
+		return se == syscall.ECONNRESET || se == syscall.ECONNREFUSED
```
broady, line 244: Need this behind a build flag.

Also, I think you can do:

```go
if err == syscall.ECONNRESET || err == syscall.ECONNREFUSED {
  // ...
}
```

no need to check the type first.

For the build flags, maybe something like

```go
// isRetryable is a platform-specific hook, specified in retryable_linux.go
var syscallRetryable func(error) bool

// ...

if syscallRetryable != nil && syscallRetryable(err) {
   return true
}
```
```go
// +build linux

func init() {
  syscallRetryable = func(err error) bool { return err == syscall.ECONNRESET || err == syscall.ECONNREFUSED }
}
```

tritone: Thanks for the suggestion! Fixed.

tritone:
Looks like there was a flake in a grpc transport test above, re-running kokoro.

tritone:
Got a 👍  offline from @broady. I'm going to go ahead and merge this.

PR #534, opened 2020-06-16
transport/grpc: support GOOGLE_API_USE_MTLS=always

broady:
This will swap "foo.googleapis.com" to "foo.mtls.googleapis.com" and used by mtls_smoketest.

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -184,7 +184,36 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 		grpcOpts = append(grpcOpts, timeoutDialerOption)
 	}
 
+	// NOTE(cbro): this is used only by the nightly mtls_smoketest and should
+	// not otherwise be used. It will be removed or renamed at some point.
+	if os.Getenv("GOOGLE_API_USE_MTLS") == "always" {
+		o.Endpoint = generateDefaultMtlsEndpoint(o.Endpoint)
+	}
+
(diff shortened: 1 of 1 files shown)


transport/grpc/dial.go
```diff
@@ -184,9 +184,37 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 		grpcOpts = append(grpcOpts, timeoutDialerOption)
 	}
 
+	// TODO(cbro): support "autodetect" and "never", and possibly merge with/refactor the equivalent in htransport.
+	if os.Getenv("GOOGLE_API_USE_MTLS") == "always" {
```
codyoss, line 188: Is this different from `GOOGLE_API_USE_CLIENT_CERTIFICATE` discussed in ADC design doc?

codyoss: Also, should this apply to HTTP as well?

broady: Already implemented in HTTP. This CL is giving the minimum support for the feature for our nightly smoke test (critical - can give more background over chat).

But yes, the semantics and name changed, so there'll be a separate implementation for GOOGLE_API_USE_CLIENT_CERTIFICATE in the future.

GOOGLE_API_USE_TLS will become only for the smoketest. I'll update the comment to reflect that.

codyoss: Thanks for the clarification.

shinfan:
Thanks for updating this Chris.

PR #537, opened 2020-06-17
gensupport: reintroduce retry deadline

tritone:
It turns out that ResumableUpload.Upload actually uses
SendRequest to send the request ultimately (via the
transferChunk and doUploadRequest methods). Therefore,
adding a retry to SendRequest effectively removes the
existing deadline for each chunk.

This commit re-adds a deadline like what was in the
original form of #528 as a safety mechanism so that callers
don't see new confusing hangs during the resumable upload.
The test that checked for indefinite retries in an upload
was also re-added.

In the future, it makes sense to either remove the retry
in ResumableUpload.Upload, or revert the changes in SendRequest
and create a new SendAndRetryRequest function instead. But
either of those (especially the first) seems complicated to
do in the immediate term. And the double retry loop we have
here is ultimately harmless because the outer loop will
immediately timeout once the inner loop is done retrying.

Fixes googleapis/google-cloud-go#2469

### DIFF
--- a/internal/gensupport/resumable_test.go
+++ b/internal/gensupport/resumable_test.go
@@ -301,6 +301,57 @@ func TestCancelUploadBasic(t *testing.T) {
 	}
 }
 
+func TestRetry_Bounded(t *testing.T) {
+	const (
+		chunkSize = 90
+		mediaSize = 300
+	)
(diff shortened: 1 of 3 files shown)


tritone:
Decided to fix the issue by submitting #538 instead, so I'm going to close this PR.

PR #538, opened 2020-06-17
gensupport: only retry storage media upload requests

### LINKED ISSUE #2469: androidpublisher: missing TargetingRuleScope fields
> **PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Filed Case 50116975 

#### Environment details
  - Programming language: golang
  - OS: linux
  - Language runtime version: go 1.21
  - Package version: v0.114.0

#### Steps to reproduce
The androidpublisher client defines target rules for subscription offers  [here](https://github.com/googleapis/google-api-go-client/blob/67f16b6ada0cd92d94d98bb32f3bec61ca3472c6/androidpublisher/v3/androidpublisher-gen.go#L8066), but only includes `SpecificSubscriptionInApp`. 

It's missing the other targeting rules defined on the developer docs for the androidpublisher API, [defined here](https://developers.google.com/android-publisher/api-ref/rest/v3/monetization.subscriptions.basePlans.offers#targetingrulescope), namely `anySubscriptionInApp` and `thisSubscription`

tritone:
This modifies the changes in #528 by reverting the changes
to SendRequest (which is widely used) and instead moving the
new logic to a separate function, SendAndRetryRequest which
will only be used on the initial request for media uploads.

There is some code duplication from SendRequest now, which I
can de-duplicate if folks think this is a good approach

Fixes googleapis/google-cloud-go#2469


broady:
CL title should probably say something like "only retry media upload requests"?

The method name is an internal detail (gensupport is internal). But if someone were to look through commit history, this would explain a change in behavior between versions (in addition to release notes of course)

broady:
er, *PR* title :)

tritone:
@broady good point, thanks. :)

google-api-go-generator/gen.go
```diff
@@ -2061,8 +2061,11 @@ func (meth *Method) generateCode() {
 		}
 		pn(`})`)
 	}
-
-	pn("return gensupport.SendRequest(c.ctx_, c.s.client, req)")
+	if meth.supportsMediaUpload() {
```
codyoss, line 2064: This will add retries for 60 different api calls, maybe we should just scope this to storage(which has this on one api) for now until retires are evaluated more across the board. Thoughts?

tritone:
Discussed this with @codyoss offline-- we do think it makes sense to limit scope to storage for right now until the effects have been more fully evaluated. It ended up being a simple change in gen.go to accomplish this (I verified by running the generation locally and confirming that only the one call I was targeting in storage-gen.go is now impacted once generation runs).

PR #541, opened 2020-06-18
all, transport/http: read default mTLS endpoint from Discovery Doc

andyrzhao:
Read default mTLS endpoint from Discovery Doc instead of generating it via regex. This is achieved by introducing a new internaloption "DefaultMTLSEndpoint". The generator is updated to include mtlsBasePath. Testdata is updated to include mtlsRootUrl.


google-api-go-generator/gen.go
```diff
@@ -495,6 +495,19 @@ func (a *API) apiBaseURL() string {
...
 
+func (a *API) mtlsApiBaseURL() string {
+	var base, rel string
+	switch {
+	case *baseURL != "":
+		base, rel = *baseURL, a.doc.BasePath
+	case a.doc.MTLSRootURL != "":
+		base, rel = a.doc.MTLSRootURL, a.doc.ServicePath
+	default:
+		base, rel = *apisURL, a.doc.BasePath
```
broady, line 506: Why does this fall-back to the default base path?

It should probably return empty string if there is none defined in the discovery doc.

google-api-go-generator/gen.go
```diff
@@ -740,7 +755,7 @@ func (a *API) GenerateCode() ([]byte, error) {
 		pn("// NOTE: prepend, so we don't override user-specified scopes.")
 		pn("opts = append([]option.ClientOption{scopesOption}, opts...)")
 	}
-	pn("opts = append(opts, internaloption.WithDefaultEndpoint(basePath))")
+	pn("opts = append(opts, internaloption.WithDefaultEndpoint(basePath), internaloption.WithDefaultMTLSEndpoint(mtlsBasePath))")
```
broady, line 758: If above suggestion is accepted, have it only add the option if `mtlsBasePath != ""`.

google-api-go-generator/gen.go
```diff
@@ -714,11 +727,13 @@ func (a *API) GenerateCode() ([]byte, error) {
 	pn("var _ = strings.Replace")
 	pn("var _ = context.Canceled")
 	pn("var _ = internaloption.WithDefaultEndpoint")
+	pn("var _ = internaloption.WithDefaultMTLSEndpoint")
```
broady, line 730: No need for this, only one symbol. is needed per package we import.

```suggestion
```

google-api-go-generator/gen.go
```diff
@@ -495,6 +495,19 @@ func (a *API) apiBaseURL() string {
 	return resolveRelative(base, rel)
 }
 
+func (a *API) mtlsApiBaseURL() string {
```
broady, line 498: Api -> API

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

google-api-go-generator/gen.go
andyrzhao: mtlsAPIBaseURL mirrors the logic in apiBaseURL, which uses the following fall-back order:
1. Check for baseURL flag override (this will be the authoritative endpoint, mtls or not, if provided)
2. If MTLSRootURL is available in discovery doc, use it.
3. Fall-back to apisURL, which is actually the discovery url "https://www.googleapis.com/discovery/v1/apis".
I'm not sure why step 3 returns discovery url instead of just empty. Either way, we should make the behavior consistent between mtls and nonMTLS. WDYT?

broady: not sure, that sounds like a bug. I think mTLS should read the field directly from the discovery doc and we can move any logic into the transport layer instead of having it in the generator.

andyrzhao: Ok, simplifying this to:
func (a *API) mtlsAPIBaseURL() string {
	if a.doc.MTLSRootURL != "" {
		return resolveRelative(a.doc.MTLSRootURL, a.doc.ServicePath)
	}
	return ""
}

google-api-go-generator/gen.go
andyrzhao: done

google-api-go-generator/gen.go
andyrzhao: done

google-api-go-generator/gen.go
andyrzhao: done

google-api-go-generator/gen.go
shinfan: I think we will still need this fallback logic though, if the url is not in the discovery doc, the transport package should fallback to the auto constructed path

andyrzhao: I think if the mTLS url is not in the discovery doc (which would be rare), then in all likelihood the service does not support mTLS anyway (and the enable_mtls flag would probably be false in the service config to begin with). In a subsequently PR, I will add the logic to incorporate the "enable_mtls" flag into the client, which will authoritatively control whether to allow mTLS endpoint in the client. See go/mtls-client-side-whitelisting.

google-api-go-generator/gen.go
```diff
@@ -741,6 +749,9 @@ func (a *API) GenerateCode() ([]byte, error) {
 		pn("opts = append([]option.ClientOption{scopesOption}, opts...)")
 	}
 	pn("opts = append(opts, internaloption.WithDefaultEndpoint(basePath))")
+	if a.mtlsAPIBaseURL() != "" {
+		pn("opts = append(opts, internaloption.WithDefaultMTLSEndpoint(mtlsBasePath))")
+	}
```
broady, line 754: keep here and L729 consistent. people *do* sometimes edit this code by hand.

suggest adding the same if statement above or change this to have the if statement in the generated code.

like this suggestion

```suggestion
pn("if mtlsBasePath != "" { opts = append(opts, internaloption.WithDefaultMTLSEndpoint(mtlsBasePath)) }")
```

google-api-go-generator/gen.go
```diff
@@ -719,6 +726,7 @@ func (a *API) GenerateCode() ([]byte, error) {
 	pn("const apiName = %q", a.doc.Name)
 	pn("const apiVersion = %q", a.doc.Version)
 	pn("const basePath = %q", a.apiBaseURL())
+	pn("const mtlsBasePath = %q", a.mtlsAPIBaseURL())
```
broady, line 729:
```suggestion
	if mtlsBase := a.mtlsAPIBaseURL(); mtlsBase != "" {
		pn("const mtlsBasePath = %q", mtlsBase)
	}
```

("opts = append(opts, internaloption.WithDefaultMTLSEndpoint(mtlsBasePath))")

andyrzhao: done

PR #557, opened 2020-06-25
transport/grpc: don't reassign ClientSettings.Endpoint

broady:
dial is called multiple times when connection pooling is used,
and reassigning o.Endpoint multiple times generated endpoints like:
pubsub.mtls.mtls.googleapis.com


broady:
This is used only for mtls_smoketest

Issue #565, opened 2020-07-02
idtoken: client created with WithCredentialsJSON failed


marvin5064:
trying to use idtoken with google service account for cloud function execution 

I got error `multiple credential options provided` from [this line](https://github.com/googleapis/google-api-go-client/blob/8809fe2216f75c184cf682992f90bd98f325c278/internal/settings.go#L76-L78)

example:
```
idtoken.NewClient(ctx, endpoint, option.WithCredentialsJSON([]byte("service account json payload")))
```

with `option.WithCredentialsFile`, it seems working fine, probably because of the [backwards compatibility](https://github.com/googleapis/google-api-go-client/blob/8809fe2216f75c184cf682992f90bd98f325c278/internal/settings.go#L76) introduced 2 years ago (which didn't count in the CredentialJSON one)

with the todo listed [here](https://github.com/googleapis/google-api-go-client/blob/8809fe2216f75c184cf682992f90bd98f325c278/internal/settings.go#L56), shall we remove other cred options from [here](https://github.com/googleapis/google-api-go-client/blob/8809fe2216f75c184cf682992f90bd98f325c278/idtoken/idtoken.go#L53), only keep a tokensource before entire next line?

codyoss:
Thanks for the report, we will get this fixed. For now, a couple workarounds:

- Write JSON to a tempfile
- Do similar logic found in NewClient, but only pass in the relevant ClientOptions

PR #570, opened 2020-07-08
fix(idtoken): allow WithCredentialsJSON to work with NewClient

codyoss:
Added new `internaloption.SkipValidation` that makes the call
to validate ClientOptions a no-op.

Fixes: #565


internal/transport/http/dial.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2020 Google LLC.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package http
```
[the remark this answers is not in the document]
codyoss: I was trying to mimic the old package structure with the refactor, but I suppose this is not necessary. I am open to suggestions though, naming is hard 😄

broady:
Hmmm, I think I get what steps led to moving things to another package but I'd rather figure out a solution that doesn't involve a refactor like this. It makes an already difficult set of packages slightly more difficult to navigate.

A couple ideas:
```
internaloption.SkipValidation()
```

Clear the conflicting fields:
```
o = append(o, option.WithCredentialJSON(nil), option.WithCredentialsFile(""), option.WithCredenoption.WithTokenSource(...))
```

Or maybe we make it easier to directly manipulate the settings struct:

```
s.Credentials = nil
s.CredentialsFile = ""
s.CredentialsJSON = nil
s.TokenSource = ...
o = []option.ClientOption{internaloption.WithSettings(s)}
```

codyoss:
> internaloption.SkipValidation()

I had considered something like at one point... Seems like a good escape hatch to me. Will refactor

internal/transport/http/dial.go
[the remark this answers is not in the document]
codyoss: No longer relevant with latest refactor.

idtoken/idtoken.go
```diff
@@ -50,7 +51,9 @@ func NewClient(ctx context.Context, audience string, opts ...ClientOption) (*htt
 	if err != nil {
 		return nil, err
 	}
-	opts = append(opts, option.WithTokenSource(ts))
+	// skipping validation so added TokenSource will not conflict with user
+	// provided credentials.
+	opts = append(opts, option.WithTokenSource(ts), internaloption.SkipValidation())
```
tbpg, line 56: Bikeshed: I don't know what `SkipValidation` means here. Perhaps something more specific? `SkipDialSettingsValidation`? It's longer, but it's more clear what's happening.

idtoken/idtoken.go
```diff
@@ -50,7 +51,9 @@ func NewClient(ctx context.Context, audience string, opts ...ClientOption) (*htt
 	if err != nil {
 		return nil, err
 	}
-	opts = append(opts, option.WithTokenSource(ts))
+	// skipping validation so added TokenSource will not conflict with user
```
tbpg, line 54:
```suggestion
	// Skip DialSettings validation so added TokenSource will not conflict with user
```

Also, it looks like `ds.Validate` has already been called above?

option/internaloption/internaloption.go
```diff
@@ -34,7 +34,19 @@ func (o defaultMTLSEndpointOption) Apply(settings *internal.DialSettings) {
 // WithDefaultMTLSEndpoint is an option that indicates the default mTLS endpoint.
 //
 // It should only be used internally by generated clients.
-//
 func WithDefaultMTLSEndpoint(url string) option.ClientOption {
 	return defaultMTLSEndpointOption(url)
 }
+
+// SkipValidation bypasses validation on ClientOptions.
```
tbpg, line 41: I'm forgetting, why don't we have an actual `internal` package for this?

codyoss: So that cloud.google.com/go can use these options as well. At least that is why the rest are here and not in `internal`

idtoken/idtoken.go
codyoss: That one is safe to call, and should be called. The validation only needs to be loosened here because when `WithTokenSource` is added, if the user had supplied other credential type options and future validations would fail.

idtoken/idtoken.go
codyoss: I think that is a fair call out. Being extra verbose here seems okay to me since this should only be used by us and promotes readability.

PR #584, opened 2020-07-14
chore(idtoken): speed up test by not using time.Sleep; use a fake clock

bradfitz:


codyoss:
Thanks for the contribution!

Issue #590, opened 2020-07-16
all: regen PRs should not need to be manually updated


codyoss:
For some reason our last couple of nightly regens are adding a dep to our `go.mod` file that is not really needed. Example #589. I have manually ran `go mod tidy` on the PR to fix it. Let's figure out why this is happening all of the sudden.

codyoss:
The reason why I could not reproduce this locally is because I use 1.14. It seem like changes to the `go` command in 1.13 limited the number of dependencies that would be auto-updated when running `go build` or `go mod download`.

codyoss:
Related:
- https://github.com/golang/go/issues/34450
- https://github.com/golang/go/issues/34822

codyoss:
This is now complete. PRs now have a static title and are generated with the same version as they are tidied with.

Issue #519, opened 2020-06-04
internal: switch automation or clean-up code


codyoss:
We might be moving away from auto-synth in the not too distant future. If we do we will need the following code for opening up pull requests. If we decide to stay on the current automation this code should be removed.

Code: https://github.com/googleapis/google-api-go-client/tree/master/internal/kokoro/synth/sendpr

codyoss:
After more investigation I think autosynth is not the right choice for us right now.

1. It is more tailored for generating gapics with context aware commits.
2. There is no ability to change the PR title. This means upon squashing the changes you need to manually rename the commit every time.
3. The PR will not update if open for more than one day.

For these reasons I think we should do something similar to what we are doing in google-cloud-go, have our own regen process.

PR #599, opened 2020-07-24
feat(internal/kokoro): switch off of autosynth onto nightly kokoro

codyoss:
Right now we use autosynth to regenerate nightly. This has been
working okay for us, but does have some downsides.

1. We can't have a programmatic naming of PRs. Right now for each
PR opened we have to manually rename the PR title each time. In
the future regens might auto-merge on green, this would be a
blocker for that.

2. We don't control the build env. Recently, none of our regens
are green from the start. They all need to have `go mod tidy` ran
from an environment of go1.13+. We could add a step in our script
to install go 1.13 everytime we regen, but it seems cleaner to
just run from one of our existing Go docker images with trampoline.

Fixes: #590
Fixes: #519


codyoss:
There is no backing job for this yet, will create one once this is reviewed.

codyoss:
Closing in favor of #649

Issue #378, opened 2019-07-12
Add ImpersonatedCredentials 


salrashid123:
Add capability to impersonate a service account.

This feature is idential to the capability existing in 
- [google-auth-python](https://google-auth.readthedocs.io/en/latest/reference/google.auth.impersonated_credentials.html)
- [google-auth-java: ImpersonatedCredentials.java](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java)

the original PR against oauth2 was abandoned on suggestion it gets added to this repo
ref: https://github.com/golang/oauth2/issues/336

as  a folwlup, i created https://github.com/googleapis/google-api-go-client/pull/337  againts this repo but on request from the maintainers here back then,  they would implement it separately somehow.   Its not implemented yet so are we ok to reopen the the PR 337?  

If so, the sample usage looks like
https://gist.github.com/salrashid123/7994be642c466844f85e2f2c68a5dec9

jadekler:
Hey @salrashid123! We're working on this, but it's going to be a while. It's moderately tricky to make this work properly, and as users expect in all the corner cases, and how it fits in with all the other credentials.

salrashid123:
until its figured out and implemented:
 an  impersonatedcredentials and a way to get idTokens.  it only works if credential object is supplied

- https://github.com/salrashid123/oauth2

Bowbaq:
@jadekler Is there any update on this. I'm running into https://issuetracker.google.com/issues/142086394 and it seems that the only fix available to me would be to impersonate a user

salrashid123:
@Bowbaq i dont' think you're after the same impersonation as requested here.  If you're aftr gsuites "delegated access" (eg, for drive, spreadsheets, etc), see

https://gist.github.com/salrashid123/7994be642c466844f85e2f2c68a5dec9#gistcomment-2968833

Bowbaq:
Actually, the reporter of that issue uses G Suite, but I don't, though I also run into the same problem with emails not being sent documents are shared by a service account.

gaahrdner:
I too, desperately need this.

shellcromancer:
+1 to adding this feature. Would be super helpful with managing IAM as Code and restricting access in GCP

salrashid123:
from a comment a bit above, i've put together `ImpersonatedCredential` TokenSource here
[...]

PR #625, opened 2020-08-12
feat(option): support service account impersonation

codyoss:
This adds support for service account impersonation with the use
of a new client option. This implementation only supports
service-account/user-credential impersonating another service
account. This does not cover the use-case that involves
impersonating a service account as an admin user with domain wide
delegation.

Fixes: #378


codyoss:
cc @salrashid123

internal/impersonate/impersonate.go
```diff
@@ -0,0 +1,127 @@
...
+		Delegates: i.delegates,
+		Lifetime:  i.lifetime,
+		Scope:     i.scopes,
+	}
+	b, err := json.Marshal(reqBody)
+	if err != nil {
+		return nil, fmt.Errorf("impersonate: unable to marshal request: %v", err)
+	}
+	url := fmt.Sprintf("https://iamcredentials.googleapis.com/v1/%s:generateAccessToken", i.name)
+	req, err := http.NewRequestWithContext(i.ctx, "POST", url, bytes.NewReader(b))
```
codyoss, line 96: Looks like this api is too new, requires 1.13. I will refactor it.

integration-tests/impersonate/impersonate_test.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2020 Google LLC.
```
broady, line 1: what is this directory "integration-tests"? does it need to be a top-level directory?

option/option.go
```diff
@@ -269,3 +270,57 @@ type withClientCertSource struct{ s ClientCertSource }
...
+// is used while not using a NewClient/NewService function, WithScopes must also
+// be explicitly passed in as well.
+//
+// If the base credential is an authorized user and not a service account, or if
+// the option WithQuotaProject is set, the target service account must have a
+// role that grants the serviceusage.services.use permission such as
+// roles/serviceusage.serviceUsageConsumer.
+//
+// This is an EXPERIMENTAL API and may be changed or removed in the future.
+func WithImpersonatedCredentials(target string, delegates ...string) ClientOption {
```
broady, line 308: Most of the time, WithX means you're passing in an X.

In this case, we're not passing in impersonated credentials (which would be a *google.Credentials type, I think) but rather named delegates.

`WithImpersonatedDelegate" perhaps?

"WithImpersonationTarget" ?

Sorry, just riffing. I don't think those are particularly good names, but that's all I've got right now.

internal/creds.go
```diff
@@ -11,13 +11,25 @@ import (
...
 	"golang.org/x/oauth2"
+	"google.golang.org/api/internal/impersonate"
 
 	"golang.org/x/oauth2/google"
 )
 
 // Creds returns credential information obtained from DialSettings, or if none, then
 // it returns default credential information.
 func Creds(ctx context.Context, ds *DialSettings) (*google.Credentials, error) {
+	creds, err := creds(ctx, ds)
```
broady, line 22: rename unexported "creds" to "baseCreds" perhaps? indicating that it doesn't cover impersonation.

internal/creds.go
```diff
@@ -103,3 +115,15 @@ func QuotaProjectFromCreds(cred *google.Credentials) string {
...
+
+func impersonateCredentials(ctx context.Context, creds *google.Credentials, ds *DialSettings) (*google.Credentials, error) {
+	if len(ds.ImpersonationConfig.Scopes) == 0 {
+		ds.ImpersonationConfig.Scopes = ds.Scopes
+	}
+	ts, err := impersonate.TokenSource(ctx, creds.TokenSource, ds.ImpersonationConfig)
+	if err != nil {
+		return nil, err
+	}
+	creds.TokenSource = ts
```
broady, line 127: wonder if creds.JSON needs to be niled out here.

broady: I think you should probably return a brand new *google.Credentials

```
return &google.Credentials{ TokenSource: ts }
```

internal/impersonate/impersonate.go
```diff
@@ -0,0 +1,128 @@
...
+	if len(config.Scopes) == 0 {
+		return nil, fmt.Errorf("impersonate: scopes must be provided")
+	}
+	its := impersonatedTokenSource{
+		ctx:  ctx,
+		ts:   ts,
+		name: formatIAMServiceAccountName(config.Target),
+		// Default to the longest acceptable value of one hour as the token will
+		// be refreshed automatically.
+		lifetime: "3600s",
```
salrashid123, line 45: coudl we make an optional "lifetime" config parameter:
 https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken

the usecase is if i want to impersonate, then get an access token, then downscope/attenuate its permissions.   I can attenuate the resources the new token can access but can't toggle its lifetime.

if i can set the total lifetime on the parent token, that will get inherited to the downscoped token

see (using-impersonated-idtokens-or-impersonated-downscoped-credentials)[https://github.com/salrashid123/oauth2#using-impersonated-idtokens-or-impersonated-downscoped-credentials]

option/option.go
codyoss: Do you like `ImpersonateCredentials`? Just drop the with?

integration-tests/impersonate/impersonate_test.go
codyoss: It already exists today. It holds some tests for idtoken and storage today. Maybe in a future PR we should move it to internal?

internal/creds.go
codyoss: Ack.

internal/creds.go
codyoss: You are probably right about the JSON, but maybe we still want the ProjectId field to be copied through?

internal/impersonate/impersonate.go
codyoss: This package is all internal right now is currently only exposed through the client option `WithImpersonatedCredentials`. There was some discussion in the design doc about having a configurable expiry, but we decided against it for now. We don't currently allow the lifetime of an of our tokens to be set. It was brought up that if we wanted to support expiry we would want to do it holistically with another client option. Also, this implementation auto-refreshes the token to make sure the TokenSource can be used w/o users having to create new impersonated credential.

If we did expose expiry, we would not want to refresh the token if set, right? Or do you think there is still valuable to have a short lifetime even though the token would be automatically refreshed? 

When/If we decide to make this package public I think such an option would be more useful.

salrashid123: - if this is implemented all as an option, the use is directly with a GCP API then, correct?   Where i'm going at is there are common enough situations where i'd like derive other credential types source from impesonated credentials. For example, get an OIDC token that represents the impersonated credential.  Or use the impersonated credential to `signBlob()` (eg, generate [IAM based SignedURL](https://github.com/googleapis/google-cloud-go/issues/1130#issuecomment-668863950))

re: lifetime
* as mentioned, its exposed in the IAM today and has a current usecase (eg, derive a downscoped token)

* re: autorefresh.  this implmentation should chain back through with the underlying tokencsource/credentials once it expires if that is what is intended by the requestor. 

* all our other implementations supports this feature as a `Credential` by itsef that allows for setting the lifetime: [python](https://google-auth.readthedocs.io/en/latest/reference/google.auth.impersonated_credentials.html), [java](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ImpersonatedCredentials.java#L107), [terraform](https://www.terraform.io/docs/providers/google/r/google_service_account.html)

i guess my biggest concern is the overall implementation (i.,e option vs as a tokensource in its own right).  Using the option fits in well w/ using this agains our APIs but there are other situations IAM based capabilities are used.

codyoss: > if this is implemented all as an option, the use is directly with a GCP API then, correct?

The primary use-case for exposing this as an option  is so it integrates nicely with any of our client libraries. For example:

```go
storage.NewClient(context.Background(), option. WithImpersonatedCredentials("foo@project.iam.gserviceaccount.com"))
```

But the option paradigm is used in other packages as well like [google.golang.org/api/transport](https://pkg.go.dev/google.golang.org/api/transport). This would allow someone to get a [*google.Credentials](https://pkg.go.dev/golang.org/x/oauth2/google?tab=doc#Credentials) from [Creds](https://pkg.go.dev/google.golang.org/api/transport?tab=doc#Creds); which would allow someone to get the impersonated [oauth2.TokenSource](https://pkg.go.dev/golang.org/x/oauth2@v0.0.0-20200107190931-bf48bf16ab8d?tab=doc#TokenSource)

> i guess my biggest concern is the overall implementation (i.,e option vs as a tokensource in its own right)

If we want to more easily empower other use-cases down the road we could look at refactoring the internal `impersonate` package and make it external. This would allow us the flexibility to also support other use cases like DwD more easily -- directly from that package.

salrashid123: yeah, i can see how its convenient for the direct with a api library but its leaving out substantial usecases pointed out in the cited bug #378 as well as reasonable consistency with the other google auth library sets.

Can you provide a concrete working example of what was cited as deriving the `TokenSource` from this implementation?

codyoss:
```go
        // import "google.golang.org/api/transport"
	creds, err := transport.Creds(context.Background(), option.WithImpersonatedCredentials("foo@project.iam.gserviceaccount.com"), option.WithScopes("https://www.googleapis.com/auth/cloud-platform"))
	if err != nil {
		log.Fatalf("problem getting creds: %v", err)
	}
	ts := creds.TokenSource
	_ = ts
```

salrashid123: ok, thats better.. so the transport would allow you to chain options and auto-refresh them
 https://godoc.org/google.golang.org/api/option#WithCredentials
i.,e you provide a serviceaccountJSON and from there derive a credential/tokensource thats impersonated.  when the derived token expires, it will look towards the source.

i'm guessing ther'es another PR somewhere to attach a similar options to derived id_tokens
   https://godoc.org/google.golang.org/api/idtoken

codyoss: > ok, thats better.. so the transport would allow you to chain options and auto-refresh them
https://godoc.org/google.golang.org/api/option#WithCredentials

Correct.

> i'm guessing ther'es another PR somewhere to attach a similar options to derived id_tokens
https://godoc.org/google.golang.org/api/idtoken

Yes, some work will have to be done there to support the option. This bring up a good point though that we should return an error for now in the `idtoken` package until such a feature is implemented.

salrashid123: ok, more thing,  if someone uses the `storage` library and its bootstrapped with the impersonated credentials objet, it will fail during signedURL operations because the library expects a real certificate to be present:

https://github.com/googleapis/google-cloud-go/blob/master/storage/storage.go#L640
  (i don't know if the storage client checks to see if the provided creds can even be used to sign though)

it gets a bit more convoluted but to support signedurl + impersonation requeres use of the IAM api to `signBlob()` again:

eg, like
  https://github.com/googleapis/google-cloud-go/issues/1130#issuecomment-668863950

for now i guess that can also throw an unsupported credential error or something

codyoss: > it will fail during signedURL operations because the library expects a real certificate to be present

The [SignedURL](https://pkg.go.dev/cloud.google.com/go/storage?tab=doc#SignedURL) stuff for storage is just a top level function that you pass options to. It does not use an authenticated client.

> to support signedurl + impersonation requeres use of the IAM api to signBlob()

Correct, in the case of impersonated credentials you would never have a PrivateKey anyways though, right? So I think you would need to take the approach of using `SignedURLOptions. SignBytes`.

option/option.go
codyoss: Done.

salrashid123:
its fine by me but it really should this is really only applicable for direct api calls to GCP services and does not address impersonated+(id_token||dwd||signedURL)

tbpg:
> this is really only applicable for direct api calls to GCP services

That makes sense to me? It's part of the http://google.golang.org/api/option package, which "contains options for Google API clients."

salrashid123:
right..semantically this PR as an option package is doing what its intended to do (so in that respect its LGTM)..but as is, the other usecases cited are simply not possible,  my point is that we'd need to account for ways to enable those in the near future (like expose the internal machinery outside of using opaque options.

tbpg:
> the other usecases cited are simply not possible

What other use cases? _If we expect this option to be able to do them_ (even if only in the future), will we be able to extend this option API design to support them?

joe-a-t:
What is the status of this since it's been 2 weeks without activity? It looks like it would be great for https://github.com/hashicorp/terraform-provider-google/issues/7064 to have this feature get released

salrashid123:
the terraform version is already there:
  https://www.terraform.io/docs/providers/google/d/service_account_access_token.html

joe-a-t:
@salrashid123 see the Workaround 2 section in https://github.com/hashicorp/terraform-provider-google/issues/7064#issue-682132558 for why the option you're suggesting doesn't work

tbpg (approved):

PR #632, opened 2020-08-20
gensupport: Allow user to provide his own buffer used for uploading

zimnx:
When uploading lots of files using ObjectStorage.Insert(), **each file** allocates his own buffer. This results in increased CPU usage and high RSS memory.

WithBuffer allows user to provide his own buffer which will be used for streaming. 

Graph below shows how many allocations were done during upload of ~30k of 10MB files:
Selection_132
With this change, there're no allocations at all on library side (pool.New is code on caller side)
Selection_135


zimnx:
@googlebot I signed it!

codyoss:
@zimnx Can you please create an issue for this PR. We like to discuss features on issues before implementing them. Thanks!

zimnx:
Added #638

zimnx:
any chance to merge it soon? Other MediaOptions are not directly used in generated clients, there are only mentions about them in comments around Media function where they can be used.

tritone:
> any chance to merge it soon? Other MediaOptions are not directly used in generated clients, there are only mentions about them in comments around Media function where they can be used.

Ahh I see I was mistaken about that-- so it will just be the manual client writer that will need to be updated for storage (and maybe bigquery).

I'm going to do a bit more discussion with my colleagues and let you know. You'd also definitely have to add a test for the new media option to https://github.com/googleapis/google-api-go-client/blob/326e17a21103f4ccf44ac1b40587ce7bcdd58b14/internal/gensupport/media_test.go#L150  (but no need to do that now, let me make sure I have approval on this direction first).

internal/gensupport/buffer.go
```diff
@@ -28,6 +28,11 @@ func NewMediaBuffer(media io.Reader, chunkSize int) *MediaBuffer {
 	return &MediaBuffer{media: media, chunk: make([]byte, 0, chunkSize)}
 }
 
+// NewMediaBuffer initializes a MediaBuffer.
+func NewMediaBufferWithBuffer(media io.Reader, chunkSize int, buffer []byte) *MediaBuffer {
+	return &MediaBuffer{media: media, chunk: buffer}
```
codyoss, line 33: There should be a fallback here to call `NewMediaBuffer` if chunksize > buffer.

codyoss: Also, before the buffer gets passed in we should reset its length. I believe the impl expects the length to be 0.

internal/gensupport/media_test.go
```diff
@@ -149,27 +149,29 @@ func TestDetermineContentType(t *testing.T) {
 
 func TestNewInfoFromMedia(t *testing.T) {
 	const textType = "text/plain; charset=utf-8"
+	chunkBuffer := make([]byte, 0)
```
codyoss, line 152: Could you add a test that will make use of the `MediaBuffer`? Maybe you could do something with `ResumableUpload`?

googleapi/googleapi.go
```diff
@@ -211,6 +211,20 @@ type MediaOption interface {
...
 }
 
+type bufferOption []byte
+
+func (bp bufferOption) setOptions(o *MediaOptions) {
+	o.Buffer = bp
+}
+
+// WithBuffer returns MediaOption which sets buffer used for media uploads.
+// It needs to be at least MinUploadChunkSize long.
```
codyoss, line 221: This is never validated. I think if the buffer is less than this it should not be used and an allocation occurs.(As it would today)

googleapi/googleapi.go
```diff
@@ -211,6 +211,20 @@ type MediaOption interface {
...
+// If used together with ChunkSize, buffer needs to be at least that long.
+// If not set, each upload will allocate its own memory.
```
codyoss, line 223: I wonder if it is worth calling out explicitly about re-using buffers only after requests complete. If someone passes in the same buffer to two concurrent calls it will lead to a data race.

internal/gensupport/buffer.go
zimnx: good idea, i had to reset buffer in caller code in order to make uploads work. Added.

internal/gensupport/media_test.go
zimnx: Sure, test case added

googleapi/googleapi.go
zimnx: Added.

googleapi/googleapi.go
zimnx: Added mention about data race.

zimnx:
Hi folks, any updates regarding merging this? :)

zimnx:
@tritone @codyoss kind reminder

danp:
👋 Regularly seeing high heap/allocations which this PR could help. Anything else needed to get this merged?

danp:
If it helps anyone: we "fixed" this by setting ChunkSize=0 on all our Writers, which disables use of the buffer entirely. Since we do our own retries elsewhere that seems fine.

codyoss:
I think this is a change we may consider in the future, but I believe the main motivation for this change was for us by the storage client. I think we should track this feature request over on https://github.com/googleapis/google-cloud-go/issues if we have not already as a storage feature request.

PR #649, opened 2020-09-03
chore(internal/kokoro): use 1.15 for regen

codyoss:


internal/kokoro/synth/synth.sh
```diff
@@ -7,6 +7,13 @@
 set -e -x
 
 cd $(dirname $0)/../../..
+
+# Install Go 1.15
+curl -o /tmp/go.tgz https://dl.google.com/go/go1.15.1.linux-amd64.tar.gz && \
+    tar -C /usr/local -xzf /tmp/go.tgz && \
```
tbpg, line 13: Is Go already installed at `/usr/local`? If so, we need to delete it first.

internal/kokoro/synth/synth.sh
```diff
@@ -7,6 +7,13 @@
 set -e -x
 
 cd $(dirname $0)/../../..
+
+# Install Go 1.15
```
tbpg, line 11: Seems like this should be updating some container image, rather than reinstalling Go?

tbpg (changes requested):

internal/kokoro/synth/synth.sh
codyoss: Done.

internal/kokoro/synth/synth.sh
codyoss: This is not run in a container today.

tbpg (approved):

PR #653, opened 2020-09-10
chore(internal/kokoro): fix Go install

codyoss:
Unable to remove currently installed verision in this environment.
Instead, install in user home and add that to PATH.

### DIFF
--- a/internal/kokoro/synth/synth.sh
+++ b/internal/kokoro/synth/synth.sh
@@ -9,10 +9,11 @@ set -e -x
 cd $(dirname $0)/../../..
 
 # Install Go 1.15
+tempdir=$(mktemp)
 curl -o /tmp/go.tgz https://dl.google.com/go/go1.15.1.linux-amd64.tar.gz &&
-    tar -C $HOME -xzf /tmp/go.tgz &&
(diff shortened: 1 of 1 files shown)


internal/kokoro/synth/synth.sh
```diff
@@ -9,11 +9,10 @@ set -e -x
 cd $(dirname $0)/../../..
 
 # Install Go 1.15
-rm -rf /usr/local/go
 curl -o /tmp/go.tgz https://dl.google.com/go/go1.15.1.linux-amd64.tar.gz &&
-    tar -C /usr/local -xzf /tmp/go.tgz &&
+    tar -C $HOME -xzf /tmp/go.tgz &&
```
tbpg, line 13: Will this mess with the default GOPATH or anything? Might be better to use another directory.

tbpg (changes requested):

internal/kokoro/synth/synth.sh
codyoss: That is a very good point.... best error on the safe side.

tbpg (approved):

PR #664, opened 2020-09-15
chore: release v0.32.0

codyoss:


tbpg (approved):
Perhaps we should get release-please running?

codyoss:
> Perhaps we should get release-please running?

Yes! I plan to work on that in the next week or so here 😸

PR #677, opened 2020-09-23
feat(transport/http): Add environment variables to conform to AIP-4114

andyrzhao:
* Add GOOGLE_API_USE_CLIENT_CERTIFICATE.
* Rename GOOGLE_API_USE_MTLS to GOOGLE_API_USE_MTLS_ENDPOINT and set default to auto. The old name GOOGLE_API_USE_MTLS will be officially deprecated.


transport/http/dial.go
```diff
@@ -286,10 +301,9 @@ func getEndpoint(settings *internal.DialSettings, clientCertSource cert.Source)
 }
 
 func getMTLSMode() string {
-	mode := os.Getenv("GOOGLE_API_USE_MTLS")
+	mode := os.Getenv("GOOGLE_API_USE_MTLS_ENDPOINT")
```
codyoss, line 304: I think we should still support the old variable here. This feature has been released and could be in use, right? I recommend honoring the new one first, and fallback to the old if not set. Then mention in the commit that we are officially deprecating the use of the old environment variable.

codyoss (changes requested):

transport/http/dial.go
andyrzhao: Good point. Done.

codyoss (approved):
Thanks!

PR #682, opened 2020-09-29
feat(transport): Add Device Certificate Authentication support to GRPC

andyrzhao:
* Refactor GetClientCertificateSource and GetEndpoint helpers into a common "dca" package to be used by both http and grpc routes.
* Moved endpoint tests to dca_test.go
* Update GetEndpoint to accept naked "host:port" as URL without merging with DefaultEndpoint. This is needed for GRPC DirectPath scenario, where "DefaultEndpoint" is likely not configured.


transport/dca/dca.go
```diff
@@ -0,0 +1,124 @@
...
+// Implications of the above logic:
+// 1. If the user specifies a non-mTLS endpoint override but client certificate is
+//    available, we will pass along the cert anyway and let the server decide what to do.
+// 2. If the user specifies an mTLS endpoint override but client certificate is not
+//    available, we will not fail-fast, but let backend throw error when connecting.
+//
+// We would like to avoid introducing client-side logic that parses whether the
+// endpoint override is an mTLS url, since the url pattern may change at anytime.
+//
+// This package is not intended for use by end developers. Use the
```
codyoss, line 24: If this is not intended for direct use please make this an internal package. I suggest `transport/internal/dca`

transport/dca/dca.go
```diff
@@ -0,0 +1,124 @@
...
+// not provided by the user.
+//
+// A nil default source can be returned if the source does not exist. Any exceptions
+// encountered while initializing the default source will be reported as client
+// error (ex. corrupt metadata file).
+//
+// Important Note: For now, the environment variable GOOGLE_API_USE_CLIENT_CERTIFICATE
+// must be set to "true" to allow certificate to be used (including user provided
+// certificates). For details, see AIP-4114.
+func GetClientCertificateSource(settings *internal.DialSettings) (cert.Source, error) {
```
codyoss, line 53: It seems like `GetClientCertificateSource` and `GetEndpoint` are called sequentially in both cases. Should these functions be an internal detail that another helper function calls

transport/dca/dca.go
```diff
@@ -0,0 +1,124 @@
...
+		return settings.DefaultEndpoint, nil
+	}
+	if strings.Contains(settings.Endpoint, "://") {
+		// User passed in a full URL path, use it verbatim.
+		return settings.Endpoint, nil
+	}
+	if settings.DefaultEndpoint == "" {
+		// If DefaultEndpoint is not configured, use the user provided endpoint verbatim.
+		// This allows a naked "host[:port]" URL to be used with GRPC Direct Path.
+		return settings.Endpoint, nil
```
codyoss, line 99: This used to return an error in the case of HTTP correct? Should that behavior be preserved?

codyoss (changes requested):

transport/dca/dca.go
andyrzhao: Done.

transport/dca/dca.go
andyrzhao: Done. I added a GetClientCertificateSourceAndEndpoint helper - that will be the only exported function.

transport/dca/dca.go
andyrzhao: You are correct, and I've put some thought here and decided that in the absence of a DefaultEndpoint, our best option really is to use the user specified endpoint verbatim, even if it is not a "full URL". Also, for HTTP case, since we expect users to use the per-service generated HTTP clients, DefaultEndpoint should always be populated.

codyoss (approved):
Thanks, LGTM

transport/dca/dca.go
codyoss: SGTM.

Issue #611, opened 2020-07-30
idtoken: Default token struct to stringify as AccessToken


grayside:
Code to mint an identity token and authenticate with a Cloud Run service presents troubleshooting challenges to developers: whether the HTTP header is wrong or IAM is misconfigured, the same HTTP 401 response is sent.

Sending the wrong token in Go is as easy as serializing the entire minted token object:

```
token, _ := s.tokenSource.Token()
req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
```

Instead of attaching just the token header value

```
token, _ := s.tokenSource.Token()
req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", token.AccessToken))
```

### Proposed

Streamline the use of the stringified token so the most important use case is implicit: add `Token.String()` to output `Token.AccessToken`. This avoids the difficult oAuth troubleshooting exercise of identifying why Cloud Run rejects what seems like a reasonable token.

codyoss:
An [oauth2.Token](https://pkg.go.dev/golang.org/x/oauth2@v0.0.0-20200107190931-bf48bf16ab8d?tab=doc#Token) is the type that is returned from the `Token` method. This itself provides a method to sign a request, see [SetAuthHeader](https://pkg.go.dev/golang.org/x/oauth2@v0.0.0-20200107190931-bf48bf16ab8d?tab=doc#Token.SetAuthHeader). Also, if a user of the idtoken package uses the [NewClient](https://pkg.go.dev/google.golang.org/api/idtoken?tab=doc#NewClient) function this header is already handled for them. 

This `oauth2.Token` type is a standard in Go. That package is not Google specific and is used by the larger community for their oauth2 needs.

codyoss:
Maybe we could address this by putting some samples right in the godoc?

grayside:
Shows me not to file a late-night issue, got the wrong repo.

Identifying to use the `AccessProperty` was not difficult from looking at the struct, this is more a case where it's easy in the course of refactoring to drop that property and it does not look wrong.

As a further consideration, in GCP terms this isn't an Access token, but as you say, this is a from a general oauth interface.

A godoc sample on idtoken.NewTokenSource would save a few minutes research, but doesn't provide the kind of footgun protection I'm suggesting in the OP. If you want a godoc on that, I can send a PR.

codyoss:
I hear what you are saying. For this issue let's add a godoc sample to help steer people in right direction.

[...]

PR #683, opened 2020-09-30
docs(idtoken): add example for setting auth header

codyoss:
Fixes: #611


tbpg (approved):

idtoken/examples_test.go
```diff
@@ -0,0 +1,26 @@
+package idtoken_test
```
tbpg, line 1: License header.

tbpg (changes requested):

idtoken/examples_test.go
codyoss: Whoops, done.

tbpg (approved):

PR #719, opened 2020-10-22
fix(transport): Fix mergeEndpoint logic to support default endpoints without scheme

andyrzhao:


transport/internal/dca/dca.go
```diff
@@ -129,11 +129,17 @@ func getMTLSMode() string {
...
 		return "", err
 	}
-	u.Host = newHost
-	return u.String(), nil
+	return strings.Replace(baseURL, u.Host, newHost, 1), nil
+}
+
+func fixScheme(baseURL string) string {
+	if !strings.Contains(baseURL, "://") {
+		return "https://" + host
```
codyoss, line 142:
```suggestion
		return "https://" + baseURL
```

transport/internal/dca/dca.go
```diff
@@ -129,11 +129,17 @@ func getMTLSMode() string {
...
+	}
+	return host
```
codyoss, line 144:
```suggestion
	return baseURL
```

codyoss (changes requested):

transport/internal/dca/dca.go
andyrzhao: my bad, last minute renaming. Fixed now.

transport/internal/dca/dca.go
```diff
@@ -129,11 +129,17 @@ func getMTLSMode() string {
...
-func mergeEndpoints(base, newHost string) (string, error) {
-	u, err := url.Parse(base)
+func mergeEndpoints(baseURL, newHost string) (string, error) {
+	u, err := url.Parse(fixScheme(baseURL))
 	if err != nil {
 		return "", err
 	}
-	u.Host = newHost
-	return u.String(), nil
+	return strings.Replace(baseURL, u.Host, newHost, 1), nil
```
codyoss, line 137: This is a bit different than what is returned before. Is this intentional? See docs for [URL.String()](https://golang.org/pkg/net/url/#URL.String) for more details.

andyrzhao: Yes, this part is intentional. The problem with using u.String() is that it will inject a "//" to the beginning of the string even if I explicitly clear the scheme (i.e. we end up with "//myhost:port/foo/bar" when the baseURL does not have a scheme) The revised logic makes our intention very explicit, which is to replace default host with the new host, and avoids idiosyncrasies with URL.String(). Let me know if you have a better solution. Thanks!

codyoss (approved):

PR #721, opened 2020-10-23
feat(transport): Add default certificate caching support

andyrzhao:
Cache and return the cached certificate as long as it has not expired. This avoids having to exec the cert provider command multiple times in the same session, especially when using a dial pool.


transport/cert/default_cert.go
```diff
@@ -106,5 +114,14 @@ func (s *secureConnectSource) getClientCertificate(info *tls.CertificateRequestI
 	if err != nil {
 		return nil, err
 	}
+	defaultSourceCachedCert = &cert
 	return &cert, nil
 }
+
+func isCertificateExpired(cert *tls.Certificate) bool {
+	parsed, err := x509.ParseCertificate(cert.Certificate[0])
```
codyoss, line 122: Does there need to be a bounds check here?

codyoss: Is there a possible race condition here? We are now storing and returning a some internal state that is guarded by a mutex internally, but not externally. References could change Certificates, right?

andyrzhao: I will add a bound check for cert.Certificate (this function is private, and I'll add a comment to say we expect the parameter to always be a non-nil cert with 1 Certificate).
Race condition shouldn't be possible on the cached cert since getClientCertificate is the only function that reads/writes the cached cert, and everything inside this function including the call to check isCertificateExpired is serialized. The returned tls.Certificate should be read-only by design, though I'm not sure how to enforce this in Golang. Let me know if you have any suggestions, or a specific counter-example in mind to illustrate your concern here. Thanks!

transport/cert/default_cert.go
```diff
@@ -31,9 +33,11 @@ const (
 )
 
 var (
-	defaultSourceOnce sync.Once
-	defaultSource     Source
-	defaultSourceErr  error
+	defaultSourceOnce            sync.Once
+	defaultSource                Source
+	defaultSourceErr             error
+	defaultSourceCachedCertMutex sync.Mutex
```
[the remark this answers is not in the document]
shinfan: +1

shinfan (approved):

codyoss (approved):

codyoss:
@andyrzhao Is this good to me merged?

andyrzhao:
Yes, please go ahead and merge. Thanks!

PR #732, opened 2020-10-30
feat(internaloption): add EnableDirectPath internaloption 

mohanli-ml:
We want to make bigtable yoshi library to attempt DirectPath by default, instead of checking the environment variable GOOGLE_CLOUD_ENABLE_DIRECT_PATH. Notice that even after this PR, the real datapath is still CFE since ACL is currently denied for all projects.

Relevant Bigtable client library PR link: https://github.com/googleapis/google-cloud-go/pull/3116.

### DIFF
--- a/internal/settings.go
+++ b/internal/settings.go
@@ -41,6 +41,7 @@ type DialSettings struct {
 	CustomClaims        map[string]interface{}
 	SkipValidation      bool
 	ImpersonationConfig *impersonate.Config
+	EnableDirectPath    bool
 
 	// Google API system parameters. For more information please read:
 	// https://cloud.google.com/apis/docs/system-parameters
--- a/option/internaloption/internaloption.go
+++ b/option/internaloption/internaloption.go
@@ -50,3 +50,18 @@ type skipDialSettingsValidation struct{}
 func (s skipDialSettingsValidation) Apply(settings *internal.DialSettings) {
 	settings.SkipValidation = true
 }
+
+// EnableDirectPath returns a ClientOption that overrides the default
+// attempt to use DirectPath.
+//
+// It should only be used internally by generated clients.
+// This is an EXPERIMENTAL API and may be changed or removed in the future.
+func EnableDirectPath(dp bool) option.ClientOption {
+	return enableDirectPath(dp)
+}
+
+type enableDirectPath bool
+
+func (e enableDirectPath) Apply(o *internal.DialSettings) {
+	o.EnableDirectPath = bool(e)
+}
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -12,7 +12,6 @@ import (
 	"crypto/tls"
 	"errors"
 	"log"
-	"os"
 	"strings"
 
 	"go.opencensus.io/plugin/ocgrpc"
@@ -138,9 +137,7 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 		// * The endpoint is a host:port (or dns:///host:port).
 		// * Credentials are obtained via GCE metadata server, using the default
 		//   service account.
-		// * Opted in via GOOGLE_CLOUD_ENABLE_DIRECT_PATH environment variable.
-		//   For example, GOOGLE_CLOUD_ENABLE_DIRECT_PATH=spanner,pubsub
-		if isDirectPathEnabled(endpoint) && isTokenSourceDirectPathCompatible(creds.TokenSource) {
+		if o.EnableDirectPath && checkDirectPathEndPoint(endpoint) && isTokenSourceDirectPathCompatible(creds.TokenSource) {
 			if !strings.HasPrefix(endpoint, "dns:///") {
 				endpoint = "dns:///" + endpoint
 			}
@@ -189,7 +186,7 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 	// point when isDirectPathEnabled will default to true, we guard it by
 	// the Directpath env var for now once we can introspect user defined
 	// dialer (https://github.com/grpc/grpc-go/issues/2795).
-	if timeoutDialerOption != nil && isDirectPathEnabled(endpoint) {
+	if timeoutDialerOption != nil && o.EnableDirectPath && checkDirectPathEndPoint(endpoint) {
 		grpcOpts = append(grpcOpts, timeoutDialerOption)
 	}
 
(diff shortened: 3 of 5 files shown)


transport/grpc/dial.go
```diff
@@ -261,15 +259,7 @@ func isDirectPathEnabled(endpoint string) bool {
 		return false
 	}
 
-	// Only try direct path if the user has opted in via the environment variable.
-	directPathAPIs := strings.Split(os.Getenv("GOOGLE_CLOUD_ENABLE_DIRECT_PATH"), ",")
```
codyoss, line 265: This seems like a breaking change for users. I think if we want to support something like this feature we still need to support the old way.

internal/settings.go
```diff
@@ -41,6 +41,7 @@ type DialSettings struct {
 	CustomClaims        map[string]interface{}
 	SkipValidation      bool
 	ImpersonationConfig *impersonate.Config
+  AttemptDirectPath bool
```
codyoss, line 44: nit: `go fmt`

codyoss (changes requested):

transport/grpc/dial.go
mohanli-ml: This should be fine since the environment variable is only used internally. More background are provided in email. Thanks for catching!

internal/settings.go
mohanli-ml: Done. Thanks!

option/option.go
```diff
@@ -324,3 +324,15 @@ func (i impersonateServiceAccount) Apply(o *internal.DialSettings) {
 	o.ImpersonationConfig.Delegates = make([]string, len(i.delegates))
 	copy(o.ImpersonationConfig.Delegates, i.delegates)
 }
+
+// WithAttemptDirectPath returns a ClientOption that overrides the default 
+// attempt to use DirectPath
+func WithAttemptDirectPath(dp bool) ClientOption {
```
codyoss, line 330: Do we expect users of this library to set this any time soon? Or is this just to be used internally for now? If internal I recommend that we move this to the `internaloption` package.

option/option.go
```diff
@@ -324,3 +324,15 @@ func (i impersonateServiceAccount) Apply(o *internal.DialSettings) {
 	o.ImpersonationConfig.Delegates = make([]string, len(i.delegates))
 	copy(o.ImpersonationConfig.Delegates, i.delegates)
 }
+
+// WithAttemptDirectPath returns a ClientOption that overrides the default 
+// attempt to use DirectPath
```
codyoss, line 329: nit: Should end in a period -- https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

option/option.go
codyoss: I think a signature like `EnableDirectPath(enable bool)` might be more descriptive. The `With` convention in this package implies that is the thing being passed in as input.

option/option.go
```diff
@@ -324,3 +324,15 @@ func (i impersonateServiceAccount) Apply(o *internal.DialSettings) {
 	o.ImpersonationConfig.Delegates = make([]string, len(i.delegates))
 	copy(o.ImpersonationConfig.Delegates, i.delegates)
 }
+
+// WithAttemptDirectPath returns a ClientOption that overrides the default
```
codyoss, line 328: Can you elaborate on what direct path is here? Should we be linking to documentation on cloud.google.com?

option/option.go
codyoss, line 329: Could we start this out as an experimental option? See `ImpersonateCredentials` above as an example.

codyoss (changes requested):

option/option.go
mohanli-ml: Use DirectPath/CFE should be transparent to users, so only Yoshi libraries will need to use this option, so I agree we can move it to `internaloption`. Also I change the name to `EnableDirectPath`.

option/option.go
mohanli-ml: Thanks for pointing out!

option/option.go
mohanli-ml: Directpath is a project aims to provide unproxied communications (i.e., without CFE) between cloud (e.g, cloud VMs) and Production (e.g. borg tasks). It is an internal feature and should be transparent to real users (i.e., users should not care the data traffic is DirectPath or CFE). So I think it should be fine without a link?

option/option.go
mohanli-ml: Added the comment.

codyoss:
Forgot to push?

option/internaloption/internaloption.go
```diff
@@ -50,3 +50,18 @@ type skipDialSettingsValidation struct{}
 func (s skipDialSettingsValidation) Apply(settings *internal.DialSettings) {
 	settings.SkipValidation = true
 }
+
+// EnableDirectPath returns a ClientOption that overrides the default
+// attempt to use DirectPath.
+//
+// It should only be used by Yoshi libraries.
```
codyoss, line 57: nit: "It should only be used internally by generated clients."

This package is better labeled experimental, so let's just follow the convention from the functions and state something like this.

transport/grpc/dial.go
```diff
@@ -135,12 +134,11 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 		}
 
 		// Attempt Direct Path only if:
+    // * Yoshi library set AttemptDirectPath option
```
codyoss, line 137: nit: remove

Let's not have anything any internal names mentioned in code.

codyoss (changes requested):
Just two small nits, other than that LGTM.

option/internaloption/internaloption.go
mohanli-ml: Done.

transport/grpc/dial.go
mohanli-ml: Done.

mohanli-ml:
Update the PR so that unit test are modified accordingly.

codyoss (approved):
LGTM

internal/settings.go
```diff
@@ -41,6 +41,7 @@ type DialSettings struct {
 	CustomClaims        map[string]interface{}
 	SkipValidation      bool
 	ImpersonationConfig *impersonate.Config
+	EnableDirectPath   bool
```
codyoss, line 44: nit: go fmt

mohanli-ml: Done.

PR #738, opened 2020-11-03
feat(internaloption): add better support for self-signed JWT

codyoss:
Added a couple of new internaloptions, WithDefaultAudience and
WithDefaultScopes, which will be used to start using self-signed
JWTs by default when certain criteria are met:

1. Authenticating with a service account.
2. User does not explicitly provide their own scopes.
3. An audience is provided -- either by default from a client or
    by the user.

In the future gapic clients will begin to pass these new
internaloptions which will enable them to authenticate with with
a JWT signed by a service account. This is a non-standard oAuth2
flow and is an optimization to save an extra network request.


internal/creds.go
```diff
@@ -90,9 +90,8 @@ func credentialsFromJSON(ctx context.Context, data []byte, endpoint string, scop
 	return cred, err
 }
 
-func selfSignedJWTTokenSource(data []byte, endpoint string, audiences []string) (oauth2.TokenSource, error) {
-	// Use the API endpoint as the default audience
-	audience := endpoint
+func selfSignedJWTTokenSource(data []byte, defaultAudience string, audiences []string) (oauth2.TokenSource, error) {
+	audience := defaultAudience
```
bshaffer, line 94: Doesn't this imply that `audiences` must always be empty?

codyoss: L100 below will override the audience if the user provided one with a client option.

tbpg (approved):

internal/creds.go
bshaffer: Ahh yes, I confused `audience` with `audiences` and thought `:=` was an array push.

option/internaloption/internaloption.go
```diff
@@ -65,3 +65,32 @@ type enableDirectPath bool
 func (e enableDirectPath) Apply(o *internal.DialSettings) {
 	o.EnableDirectPath = bool(e)
 }
+
+// WithDefaultAudience returns a ClientOption that specifies a default audience
+// to be used as the audience field ("aud") for the JWT token authentication.
+//
+// It should only be used internally by generated clients.
+func WithDefaultAudience(audience string) option.ClientOption {
+	return withDefaultAudience(audience)
```
bshaffer, line 74: FYI the default audience *should* be derived from the Service Name (e.g. `https://SERVICE_NAME/`)

bshaffer (approved):

option/internaloption/internaloption.go
codyoss: It will be derived at client library generation time. This method just facilitates passing the information here.

bendiknesbo:
@bshaffer @tbpg @codyoss 
This PR broke our builds!

```go
(ds.DefaultAudience != "" || len(ds.Audiences) > 0)
```

The above change appears to be the faulting change.

Using `cloud.google.com/go v0.72.0` + `google.golang.org/api v0.35.0` works. Upgrading either with 1 minor version causes runtime errors.

Reproducing code:
```go
opt := option.WithCredentialsFile("path/to/svc-acc.json")
validator, err := idtoken.NewValidator(ctx, opt)
validator.Validate(ctx, "some-token", "some-audience")
```

The fault happens during `idtoken.NewValidator`, where the self-signing JWT token source is not set, due to the changes in this PR. There is no error returned from running `NewValidator`. When running `validator.Validate`, an error occurs during the roundtrip when retrieving the certificates from google.

codyoss:
@bendiknesbo Thanks for the report, I will look into this and get a patch out.

Issue #752, opened 2020-11-18
drive: FilesGetCall.Download did not unmarshal the body in the error response


aofei:
Just now I happened to notice that the [`FilesGetCall.Download`](https://github.com/googleapis/google-api-go-client/blob/84df58a8914832693858b396c6099dee1756bf13/drive/v3/drive-gen.go#L7050) threw me this error:

```bash
googleapi: got HTTP response code 403 with body: {"error":{"errors":[{"domain":"usageLimits","reason":"userRateLimitExceeded","message":"User Rate Limit Exceeded. Rate of requests for user exceed configured project quota. You may consider re-evaluating expected per-user traffic to the API and adjust project quota limits accordingly. You may monitor aggregate quota usage and adjust limits in the API Console: https://console.developers.google.com/apis/api/drive.googleapis.com/quotas?project=xxxxxxxxxxxxxx","extendedHelp":"https://console.developers.google.com/apis/api/drive.googleapis.com/quotas?project=xxxxxxxxxxxxxx"}],"code":403,"message":"User Rate Limit Exceeded. Rate of requests for user exceed configured project quota. You may consider re-evaluating expected per-user traffic to the API and adjust project quota limits accordingly. You may monitor aggregate quota usage and adjust limits in the API Console: https://console.developers.google.com/apis/api/drive.googleapis.com/quotas?project=xxxxxxxxxxxxxx"}}
```

The returned `error` is still [`*googleapi.Error`](https://pkg.go.dev/google.golang.org/api/googleapi#Error), which is fine. But in its fields, only `Code` and `Body` have values. This caused me to be unable to handle errors through the `Errors` field as usual.

So, does this work as designed?

tbpg:
Thanks for the issue. I'm not sure why that's the case for this call. Here is what appears to be the relevant function:
https://github.com/googleapis/google-api-go-client/blob/84df58a8914832693858b396c6099dee1756bf13/googleapi/googleapi.go#L165-L178

Note that it doesn't try to parse the body as JSON, per the function comment. It seems that all `"media"` requests use `CheckMediaResponse`, which makes sense based on naming! See https://github.com/googleapis/google-api-go-client/search?q=checkmediaresponse.

Error parsing came up for #473. So, let's wait for @codyoss. Notably, https://google.aip.dev/193 doesn't call out media requests as different.

[...]

PR #773, opened 2020-12-07
fix(all): use CheckResponse for media downloads

codyoss:
If a method supports media downloads use CheckResponse. This method
provided extra context to errors if the api follow
https://google.aip.dev/193. The storage api will continue to use
CheckMediaResponse as it does not conform to the aip due to legacy
reasons.

Fixes: #752


codyoss:
Ran the generator on cached discovery docs to generate the changes.

tbpg (commented):
Just to confirm, do we have sufficient test coverage to be confident these are still working properly?

codyoss:
Yup, everything should work just fine. We could probably get away with removing CheckMediaResponse all together in the future. `CheckResponse`'s only real difference is it will try to `json.Unmarshall` the value into something more descriptive. If the unmarshalling fails it falls back to the old behavior. At worst you lose some performance on error cases.

tbpg (approved):

PR #781, opened 2020-12-09
fix(transport/grpc): check Compute Engine environment for DirectPath

mohanli-ml:
Add Compute Engine environment check since DirectPath can only be used on Compute Engine.


transport/grpc/dial.go
```diff
@@ -186,7 +193,7 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 	// point when isDirectPathEnabled will default to true, we guard it by
 	// the Directpath env var for now once we can introspect user defined
 	// dialer (https://github.com/grpc/grpc-go/issues/2795).
-	if timeoutDialerOption != nil && o.EnableDirectPath && checkDirectPathEndPoint(endpoint) {
+	if timeoutDialerOption != nil && o.EnableDirectPath && checkDirectPathEndPoint(endpoint) && isOnComputeEngine() {
```
codyoss, line 196: Please use [metadata.OnGCE](https://pkg.go.dev/cloud.google.com/go/compute/metadata#OnGCE)

codyoss (changes requested):

transport/grpc/dial.go
mohanli-ml: Good to know!

codyoss (approved):

Issue #775, opened 2020-12-07
transport/bytestream: TestClientWrite_CloseAndRecvFails is flaky


codyoss:
Fix this test.

PR #787, opened 2020-12-11
feat(transport/bytestream): Add Close method for shutdown

codyoss:
This pacakge is still labeled expirmental. At some point we should
decide if we want to keep supporting it or deprecate and remove.

Fixes: #775


codyoss:
Ran `go test -run TestClientWrite_AfterSetupClose -count 10000` after change and all passed locally.

codyoss:
@tbpg Could you take a quick look at this when you have a chance.

tbpg (approved):

PR #788, opened 2020-12-11
fix(internal): don't use self-signed JWT with impersonation

codyoss:
The limited audience of the base credential will could cause
failures as it would not be able to comunicate with the iam
credentials api if just default scopes were passed.


internal/creds.go
```diff
@@ -64,14 +64,18 @@ const (
 
 // credentialsFromJSON returns a google.Credentials based on the input.
 //
-// - If the JSON is a service account and no scopes provided, returns self-signed JWT auth flow
-// - Otherwise, returns OAuth 2.0 flow.
+// - If the JSON is a service account, no user are scopes provided, an audience
+//   is provided, and credentials will not be impersonated executes a
```
tbpg, line 68: Missing a word before executes?

internal/creds.go
```diff
@@ -64,14 +64,18 @@ const (
...
+// - Otherwise, executes a stanard OAuth 2.0 flow.
 func credentialsFromJSON(ctx context.Context, data []byte, ds *DialSettings) (*google.Credentials, error) {
 	cred, err := google.CredentialsFromJSON(ctx, data, ds.GetScopes()...)
 	if err != nil {
 		return nil, err
 	}
-	if len(data) > 0 && len(ds.Scopes) == 0 && (ds.DefaultAudience != "" || len(ds.Audiences) > 0) {
+	if len(data) > 0 && len(ds.Scopes) == 0 && 
+	(ds.DefaultAudience != "" || len(ds.Audiences) > 0) && 
+	ds.ImpersonationConfig == nil {
```
tbpg, line 78: This is getting pretty complicated. Maybe we should invert the conditions and return early instead? Possibly split into more than one `if` statement.

tbpg (changes requested):

internal/creds.go
codyoss: rephrased.

internal/creds.go
codyoss: Inverted and split up lines to try to improve readability.

codyoss:
@tbpg Could you take a quick look at this when you have a chance.

internal/creds.go
```diff
@@ -64,28 +64,37 @@ const (
 
 // credentialsFromJSON returns a google.Credentials based on the input.
 //
-// - If the JSON is a service account and no scopes provided, returns self-signed JWT auth flow
-// - Otherwise, returns OAuth 2.0 flow.
+// - If the JSON is a service account, no user are scopes provided, an audience
+//   is provided, and credentials will not be impersonated a signed JWT auth
```
tbpg, line 68: Missing a word here?

tbpg (changes requested):

tbpg (approved):

Issue #712, opened 2020-10-19
compute: misformatted comment for Operation


dowgird:
This is the docstring generated for Operation:
```
// *
// [Global](/compute/docs/reference/rest/{$api_version}/globalOperations)
//  *
// [Regional](/compute/docs/reference/rest/{$api_version}/regionOperation
// s) *
// [Zonal](/compute/docs/reference/rest/{$api_version}/zoneOperations)
//
// Y
// ou can use an operation resource to manage asynchronous API requests.
```

It's weirdly formatted and renderss badly in HTML, for example here: https://pkg.go.dev/google.golang.org/api@v0.33.0/compute/v1#Operation

I'm referring to compute/v0.alpha/compute-gen.go, PR #707

codyoss:
Thanks for pointing this out. Appears to be the same issue as #544. I will track the progress of the issue here and close other issues. Also related to #457.

PR #789, opened 2020-12-11
fix: don't use markdown style links

codyoss:
Fixes: #712


codyoss:
Another option here is we could turn something like `This is [really important](http://example.com/important)` into `This is really important(See http://example.com/important)`. Thoughts?

codyoss:
Note manual changes are only in gen.go and gen_test.go.

tbpg (commented):
I think the links are useful. So, it seems sad to just remove them?

codyoss:
I hear you, WDYT about the "( See http://example.com approach)"?

tbpg:
Could we do `This is really important (http://example.com/important) stuff.`?

tbpg (approved):

PR #797, opened 2020-12-14
fix(transport/http): unable to supply both TokenSource and impersonated credentials

### LINKED ISSUE #796: internal: can't use impersonate service account credentials with TokenSource
With external OIDC workload identity, for example EKS OIDC, we can't use the impersonated service account token.
```go
type oidcTokenSource struct {
	projectID  string
	poolID     string
	providerID string
	tokenFile  string // "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"
}

func (ts oidcTokenSource) Token() (*oauth2.Token, error) {
	tokenData, err := ioutil.ReadFile(ts.tokenFile)
	if err != nil {
		return nil, err
	}
	ctx := context.Background()
	stsSvc, err := sts.NewService(ctx, option.WithoutAuthentication())
	if err != nil {
		return nil, err
	}
	audience := fmt.Sprintf("//iam.googleapis.com/projects/%s/locations/global/workloadIdentityPools/%s/providers/%s", ts.projectID, ts.poolID, ts.providerID)
	req := sts.GoogleIdentityStsV1betaExchangeTokenRequest{
		Audience:           audience,
		Scope:              "https://www.googleapis.com/auth/cloud-platform",
		GrantType:          "urn:ietf:params:oauth:grant-type:token-exchange",
		RequestedTokenType: "urn:ietf:params:oauth:token-type:access_token",
		SubjectToken:       string(tokenData),
		SubjectTokenType:   "urn:ietf:params:oauth:token-type:jwt",
	}
	resp, err := stsSvc.V1beta.Token(&req).Do()
	if err != nil {
		return nil, err
	}
	dur := time.Duration(resp.ExpiresIn) * time.Second
	expiry := time.Now().Add(dur)
	return &oauth2.Token{
		AccessToken: resp.AccessToken,
		TokenType:   resp.TokenType,
		Expiry:      expiry,
	}, nil
}

func main() {
	ts := oidcTokenSource{
		projectID:  projectID,
		poolID:     poolID,
		providerID: providerID,
		tokenFile:  os.Getenv("GCP_OIDC_ID_TOKEN"),
	}

	impersonateSA := os.Getenv("GCP_IMPERSONATE_SA")
	opts := []option.ClientOption{
		option.WithTokenSource(oidcTokenSource),
		option.ImpersonateCredentials(impersonateSA),
	}
	ctx := context.Background()
	client, err := storage.NewClient(ctx, opts...)
    // work with GCS client ...
}
```
With the above code snippet, I found that the client actually uses the `oidcTokenSource` instead of the impersonated service account's temporary token. Since the oidcTokenSource does not have the GCS permission, the client will report the following error:

```
panic: googleapi: Error 403: Primary: /namespaces/projects/{projectId}/locations/global/workloadIdentityPools/{poolId} with additional claims does not have storage.objects.list access to the Google Cloud Storage bucket., forbidden
```

After digging it for a while, I found this is caused by the following code:

https://github.com/googleapis/google-api-go-client/blob/76c07cb9bfecfffa076e7c107085bea68b0aa9bd/transport/http/dial.go#L82-L97

The `internal.Creds` already gets the impersonated service account's temporary credentials, but since the `oidcTokenSource` is not nil, https://github.com/googleapis/google-api-go-client/blob/76c07cb9bfecfffa076e7c107085bea68b0aa9bd/transport/http/dial.go#L91 sets the client with `oidcTokenSource` instead of the impersonated service account token source.

tennix:
This PR fixes #796


tennix:
@googlebot I signed it!

codyoss (approved):
Thank you for your contribution. This looks good to me, good catch!

PR #818, opened 2021-01-07
docs: add examples on how to test libraries

codyoss:
This is the based off the doc added to the gapic libraries. Because
of the design of the api and the fact these clients use HTTP and
not gRPC it is useful to see how testing is done here as well.

Related: googleapis/google-cloud-go#2921


internal/examples/fake/fake.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// TranslateText translates text to the given language using the provided
+// service.
+func TranslateText(service *translate.Service, text, language string) (string, error) {
+	parent := fmt.Sprintf("projects/%s/locations/global", os.Getenv("GOOGLE_CLOUD_PROJECT"))
+	resp, err := service.Projects.Locations.TranslateText(parent, &translate.TranslateTextRequest{
+		TargetLanguageCode: language,
+		Contents:           []string{text},
+	}).Do()
```
tbpg, line 23: Consider a separate `req` variable. Sometimes, it can be hard to follow multi-line function calls + declarations.

testing.md
```diff
@@ -0,0 +1,192 @@
+# Testing Code that depends on Go Client Libraries
```
tbpg, line 1:
```suggestion
# Testing Code that depends on google.golang.org/api
```

tbpg (approved):

internal/examples/fake/fake.go
codyoss: Done

testing.md
codyoss: Done

PR #843, opened 2021-01-26
feat(all): auto-regenerate discovery clients

JuriBurakov:


tbpg:
While we appreciate the PRs, we really like to discuss changes like this in an issue first. These tools are usually fully automated.

PR #847, opened 2021-01-29
fix(internal): don't self-sign JWT when an endpoint provided

codyoss:
This was recently caught by the spanner team. This is a part of the spec.


codyoss:
cc @olavloite

tbpg (approved):

bshaffer (approved):

PR #852, opened 2021-02-01
fix(transport): expand OS environment variables in cert provider command

andyrzhao:
Environment variables in the cert provider command such as "$HOME" needs to be expanded. This logic was accidentally removed in an earlier patch, so it is being added back now. It impacts scenarios using customized context_aware_metadata.json.


codyoss (approved):
LGTM

PR #857, opened 2021-02-05
oauth: Support requesting Userinfo by AccessToken

AdamGlazerMW:
This change proposes to make it possible to retrieve Userinfo by AccessToken


AdamGlazerMW:
@googlebot I signed it!

PR #858, opened 2021-02-05
feat(oauth2): add AccessToken support for UserinfoGetCall

AdamGlazerMW:
This PR suggests to extend the UserinfoGetCall with an AccessToken function which enables the user to request the Userinfo by providing the AccessToken as part of the query parameters.

### DIFF
--- a/oauth2/v1/oauth2-gen.go
+++ b/oauth2/v1/oauth2-gen.go
@@ -466,6 +466,13 @@ func (r *UserinfoService) Get() *UserinfoGetCall {
 	return c
 }
 
+// AccessToken sets the optional parameter "access_token": The oauth2
+// access token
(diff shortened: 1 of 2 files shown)


codyoss:
Hey,

Unfortunately this is generated code and can't be updated directly. Clients should already be authenticated though so there should be no reason to manually set access_token. Authentication that is added automatically happens with the Authorization header.

AdamGlazerMW:
Hi @codyoss 

My use-case would have been to verify in a separate back-end service, whether a user successfully authenticated themselves via a Chrome extension with an e-mail account that is from the company's GSuite hosted domain. But I can actually use the tokenCall one for that, and just check the email suffix. Thanks for the explanation though!

PR #881, opened 2021-02-18
chore: update oauth2 dependency

codyoss:


codyoss:
@tbpg It seems renovate has not been running in either of the libraries repos lately... I will look into this more, but for now I would like to get this dependency bumped

codyoss:
cc @bojeil-google

tbpg (approved):

PR #914, opened 2021-03-08
chore: update generator to use 1.16

codyoss:


internal/kokoro/synth/synth.sh
```diff
@@ -8,9 +8,9 @@ set -e -x
 
 cd $(dirname $0)/../../..
 
-# Install Go 1.15
+# Install Go 1.16
 tempdir=$(mktemp -d)
-curl -o /tmp/go.tgz https://dl.google.com/go/go1.15.1.linux-amd64.tar.gz &&
+curl -o /tmp/go.tgz https://dl.google.com/go/go1.16.linux-amd64.tar.gz &&
```
tbpg, line 13: Can we do something like [this](https://team-review.git.corp.google.com/c/cloud-sshrelay/dev-con/+/876966/1/cloudshell/Dockerfile) to avoid having to update this in the future? Or do we want more control for when it changes?

tbpg (approved):

Issue #777, opened 2020-12-07
idtoken: Impersonation support


mterwill:
#625 added support for impersonating service accounts where clients authenticate with an access token. It does not support ID tokens, generated with [`projects.serviceAccounts.generateIdToken`](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateIdToken). The `idtoken` package accordingly does not support impersonation:

https://github.com/googleapis/google-api-go-client/blob/c9160e8cf25085a29a4d3267e1b6ee5ad961c17c/idtoken/idtoken.go#L81-L83

I would like to use my application default credentials to impersonate a service account to authenticate to Cloud Run. Any plans to support impersonation in the `idtoken` package?

codyoss:
Yes, this is something that I would like to. The work has just not been prioritized yet. I will update this issue when there are plans to add this support in.

mterwill:
@codyoss I took a pass at implementing this!

codyoss:
Thank you for your interests in this feature! I am going to hold off on reviewing the PR for now though as we are still figuring out the impersonation story across the various languages we support for our clients today, this is one reason the feature is in an experimental status today. I will post back here when I have a better picture of what we would like to support. 

In the meantime though you still should be able to do this style of impersonation today, it just takes a little more work. Use the impersonation option to create an [iamcredentials.Service](https://pkg.go.dev/google.golang.org/api@v0.36.0/iamcredentials/v1#NewService). Then you can use the [GenerateIdToken](https://pkg.go.dev/google.golang.org/api@v0.36.0/iamcredentials/v1#ProjectsServiceAccountsService.GenerateIdToken) to get an idtoken. Hope that helps in the meantime.

mterwill:
Sounds good, thanks!

codyoss:
Thanks for understanding!

codyoss:
Please try out our impersonate preview and feel free to provide any feedback: https://github.com/googleapis/google-api-go-client/releases/tag/v0.44.0-impersonate-preview

mterwill:
@codyoss – works great, thanks! Any idea when the new package will be generally released?

codyoss:
@mterwill  Planning on cutting a release with this next week. Can follow https://github.com/googleapis/google-api-go-client/pull/991 for progress. A release will be cut shortly after the cherry pick.

mterwill:
Thank you!

codyoss:
[...]

Issue #731, opened 2020-10-29
option: Unclear best practice and documentation for option.ImpersonateCredentials


joshk0:
In #625, this comment was added to better explain the role of scopes in impersonation for the new option.ImpersonateCredentials feature.

```
// The resulting impersonated credential will either have the default scopes of
// the client being instantiating or the scopes from WithScopes if provided.
// Scopes are required for creating impersonated credentials, so if this option
// is used while not using a NewClient/NewService function, WithScopes must also
// be explicitly passed in as well.
```

Despite this comment, in my code, if I do this:

```go
 bq, err := bigquery.NewClient(ctx, c.accessProjectID, option.ImpersonateCredentials(email))
```

and then try to use `bq`, I always get the error of a 403 from the call to [generateAccessToken](https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken) and insufficient scopes described as the cause at [this line](https://github.com/googleapis/google-api-go-client/blob/ba82a99e06960678387a0b6c6ef01a0ed14c202a/internal/impersonate/impersonate.go#L113).

For the same code to work, I have to do this:

```go
 bq, err := bigquery.NewClient(ctx, c.accessProjectID, option.WithScopes("https://www.googleapis.com/auth/cloud-platform"), option.ImpersonateCredentials(email))
```

Now I can use `bq` as the impersonated service account as I expect.

I am raising this issue because the documentation seems unclear with regards to the need to pass the scopes in even though I do happen to be inside a NewClient() call. Also, it seems concerning that this will force the resulting token to be too powerful, even though I am creating it just for BigQuery purposes.

It seems like the ideal behavior would be that a native token with iam scopes is requested to perform the impersonation, then the regular requested scopes for the client or service are requested for the impersonated token. This way, tokens will be minimally scoped.

In summary I am proposing one of these outcomes:

* It is properly documented that WithScopes() must be passed to get ImpersonateCredentials() to work intuitively.
* A token with the proper iam or cloud-platform scopes is requested only for the purpose of generating tokens; the client's default scopes are then requested for the impersonated token. In other words, the call to WithScopes() shouldn't affect impersonation.

I'm also happy to be hit with the cluebat in case I'm not understanding the documentation or interface correctly. Do at least consider it a vote that the usage as documented is unclear.

cc @salrashid123 as I only knew the hint about WithScopes by reading his helpful repo, salrashid123/oauth2 👍🏼

[...]

Issue #652, opened 2020-09-09
support impersonating admin users for gSuite APIs


codyoss:
Followup tracking issue to #378.

carnott-snap:
Is it possible to implement this feature using default credentials without minting subtokens like I do in `git.sr.ht/~urandom/dwd`?

codyoss:
@carnott-snap When you say "without minting subtokens" what do you mean?

carnott-snap:
I use environments where only the application default credentials are accessable, so no json service account keys. If you look at the linked library, I achieve gsuite admin impersonation via making a new `oauth2.Token` based on the response from `cloud.google.com/go/iam/credentials/apiv1.IamCredentialsClient.SignJwt`. This new token correctly impersonates the admin, and I can confirm works as expected in production environments, but setting up the rpc and refreshing the token adds quite a bit of complexity and some runtime overhead. Furthermore, my api does not play nice with delegation and impersonation, e.g. one@appspot.gserviceaccount.com -deledates-> two@appspot.gserviceaccount.com -impersonates-> admin@example.com.

codyoss:
If the question is, would we support application default credentials(ADC) for sourcing the base credentials -- the answer is yes.

codyoss:
Now in a main release: https://github.com/googleapis/google-api-go-client/releases/tag/v0.46.0

PR #927, opened 2021-03-15
feat: add impersonate pacakge

codyoss:
This package is used for creating a impersonated TokenSource. This
TokenSource can then be passed into any client or API that accepts
a client option. This package provides support for three types of
impersonation:

1. A service account impersonating another service account.
2. A service account impersonating another service account as an
   admin user -- a pattern used with domain wide delegation.
3. A service account creating an impersonated ID token.

Fixes: #652
Fixes: #777
Fixes: #731
Updates: #378

### DIFF
--- a/impersonate/example_test.go
+++ b/impersonate/example_test.go
@@ -19,7 +19,7 @@ func ExampleTokenSource_serviceAccount() {
 	ctx := context.Background()
 
 	// Base credentials sourced from ADC or provided client options.
-	ts, err := impersonate.TokenSource(ctx, impersonate.Config{
+	ts, err := impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
 		TargetPrincipal: "foo@project-id.iam.gserviceaccount.com",
 		Scopes:          []string{"https://www.googleapis.com/auth/cloud-platform"},
 		// Optionally supply delegates.
@@ -29,7 +29,7 @@ func ExampleTokenSource_serviceAccount() {
 		log.Fatal(err)
 	}
 
-	// Pass an impersonated credential to any function that takes a client
+	// Pass an impersonated credential to any function that takes client
 	// options.
 	client, err := secretmanager.NewService(ctx, option.WithTokenSource(ts))
 	if err != nil {
@@ -45,7 +45,7 @@ func ExampleTokenSource_adminUser() {
 	ctx := context.Background()
 
 	// Base credentials sourced from ADC or provided client options.
-	ts, err := impersonate.TokenSource(ctx, impersonate.Config{
+	ts, err := impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
 		TargetPrincipal: "foo@project-id.iam.gserviceaccount.com",
 		Scopes:          []string{"https://www.googleapis.com/auth/cloud-platform"},
 		// Optionally supply delegates.
@@ -57,7 +57,7 @@ func ExampleTokenSource_adminUser() {
 		log.Fatal(err)
 	}
 
-	// Pass an impersonated credential to any function that takes a client
+	// Pass an impersonated credential to any function that takes client
 	// options.
 	client, err := admin.NewService(ctx, option.WithTokenSource(ts))
 	if err != nil {
@@ -84,7 +84,7 @@ func ExampleIDTokenSource() {
 		log.Fatal(err)
 	}
 
-	// Pass an impersonated credential to any function that takes a client
+	// Pass an impersonated credential to any function that takes client
 	// options.
 	client, _, err := transport.NewHTTPClient(ctx, option.WithTokenSource(ts))
 	if err != nil {
--- a/impersonate/example_test.go
+++ b/impersonate/example_test.go
@@ -1,0 +1,97 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package impersonate_test
+
+import (
+	"context"
+	"log"
+
+	admin "google.golang.org/api/admin/directory/v1"
+	"google.golang.org/api/impersonate"
+	"google.golang.org/api/option"
+	"google.golang.org/api/secretmanager/v1"
+	"google.golang.org/api/transport"
+)
+
+func ExampleTokenSource_serviceAccount() {
+	ctx := context.Background()
+
+	// Base credentials sourced from ADC or provided client options.
+	ts, err := impersonate.TokenSource(ctx, impersonate.Config{
+		TargetPrincipal: "foo@project-id.iam.gserviceaccount.com",
+		Scopes:          []string{"https://www.googleapis.com/auth/cloud-platform"},
+		// Optionally supply delegates.
+		Delegates: []string{"bar@project-id.iam.gserviceaccount.com"},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Pass an impersonated credential to any function that takes a client
+	// options.
+	client, err := secretmanager.NewService(ctx, option.WithTokenSource(ts))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Use your client that is authenticated with impersonated credentials to
+	// make requests.
+	client.Projects.Secrets.Get("...")
+}
+
+func ExampleTokenSource_adminUser() {
+	ctx := context.Background()
+
+	// Base credentials sourced from ADC or provided client options.
+	ts, err := impersonate.TokenSource(ctx, impersonate.Config{
+		TargetPrincipal: "foo@project-id.iam.gserviceaccount.com",
+		Scopes:          []string{"https://www.googleapis.com/auth/cloud-platform"},
+		// Optionally supply delegates.
+		Delegates: []string{"bar@project-id.iam.gserviceaccount.com"},
+		// Specify user to impersonate
+		Subject: "admin@example.com",
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Pass an impersonated credential to any function that takes a client
+	// options.
+	client, err := admin.NewService(ctx, option.WithTokenSource(ts))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Use your client that is authenticated with impersonated credentials to
+	// make requests.
+	client.Groups.Delete("...")
+}
+
+func ExampleIDTokenSource() {
+	ctx := context.Background()
+
+	// Base credentials sourced from ADC or provided client options.
+	ts, err := impersonate.IDTokenSource(ctx, impersonate.IDTokenConfig{
+		Audience:        "http://example.com/",
+		TargetPrincipal: "foo@project-id.iam.gserviceaccount.com",
+		IncludeEmail:    true,
+		// Optionally supply delegates.
+		Delegates: []string{"bar@project-id.iam.gserviceaccount.com"},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Pass an impersonated credential to any function that takes a client
+	// options.
+	client, _, err := transport.NewHTTPClient(ctx, option.WithTokenSource(ts))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Use your client that is authenticated with impersonated credentials to
+	// make requests.
+	client.Get("http://example.com/")
+}
--- a/impersonate/example_test.go
+++ b/impersonate/example_test.go
@@ -15,7 +15,7 @@ import (
 	"google.golang.org/api/transport"
 )
 
-func ExampleTokenSource_serviceAccount() {
+func ExampleCredentialsTokenSource_serviceAccount() {
 	ctx := context.Background()
 
 	// Base credentials sourced from ADC or provided client options.
@@ -41,7 +41,7 @@ func ExampleTokenSource_serviceAccount() {
 	client.Projects.Secrets.Get("...")
 }
 
-func ExampleTokenSource_adminUser() {
+func ExampleCredentialsTokenSource_adminUser() {
 	ctx := context.Background()
 
 	// Base credentials sourced from ADC or provided client options.
--- a/impersonate/idtoken.go
+++ b/impersonate/idtoken.go
@@ -1,0 +1,137 @@
+// Copyright 2021 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package impersonate
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"golang.org/x/oauth2"
+	"google.golang.org/api/option"
+	htransport "google.golang.org/api/transport/http"
+)
+
+// IDTokenConfig for generating an impersonated ID token.
+type IDTokenConfig struct {
+	// Audience is the `aud` field for the token, such as an API endpoint the
+	// token will grant access to. Required.
+	Audience string
+	// TargetPrincipal is the email address of the service account to
+	// impersonate. Required.
+	TargetPrincipal string
+	// IncludeEmail includes the service account's email in the token. The
+	// resulting token will include both an `email` and `email_verified`
+	// claim.
+	IncludeEmail bool
+	// Delegates are the service account email addresses in a delegation chain.
+	// Each service account must be granted roles/iam.serviceAccountTokenCreator
+	// on the next service account in the chain. Optional.
+	Delegates []string
+}
+
+// IDTokenSource creates an impersonated TokenSource that returns ID tokens
+// configured with the provided config and using credentials loaded from
+// Application Default Credentials as the base credentials. The tokens provided
+// by the source are valid for one hour and are automatically refreshed.
+func IDTokenSource(ctx context.Context, config IDTokenConfig, opts ...option.ClientOption) (oauth2.TokenSource, error) {
+	if config.Audience == "" {
+		return nil, fmt.Errorf("impersonate: an audience must be provided")
+	}
+	if config.TargetPrincipal == "" {
+		return nil, fmt.Errorf("impersonate: a target service account must be provided")
+	}
+
+	clientOpts := append(defaultClientOptions(), opts...)
+	client, _, err := htransport.NewClient(ctx, clientOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	its := impersonatedIDTokenSource{
+		client:          client,
+		targetPrincipal: config.TargetPrincipal,
+		audience:        config.Audience,
+		includeEmail:    config.IncludeEmail,
+	}
+	its.delegates = make([]string, len(config.Delegates))
+	for i, v := range config.Delegates {
+		its.delegates[i] = formatIAMServiceAccountName(v)
+	}
+	return oauth2.ReuseTokenSource(nil, its), nil
+}
+
+type generateIDTokenRequest struct {
+	Audience     string   `json:"audience"`
+	IncludeEmail bool     `json:"includeEmail"`
+	Delegates    []string `json:"delegates,omitempty"`
+}
+
+type generateIDTokenResponse struct {
+	Token string `json:"token"`
+}
+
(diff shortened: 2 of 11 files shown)


codyoss:
Still doing a little more manual testing for one of the use-cases, but I thought I would get some code out to start the review process as this a decent size PR.

codyoss:
Note this is being targeted to a release branch where this will be released as a preview.

codyoss:
cc @salrashid123

salrashid123:
hi- 

is this PR to break and add additional capabilities to the impersonated creds (eg, impersonate user via domain-delegation, etc) by breaking it out to a config and then to a TokenSource it its own right?

i ask since the baseline exisiting implementation allows you to derive the tokensource via `option`.


thx

---

```golang
import (
    "google.golang.org/api/option"
    "google.golang.org/api/transport"
)

func main() {

    ctx := context.Background()
    creds, err := transport.Creds(ctx,
        option.WithScopes("https://www.googleapis.com/auth/cloud-platform"),
        option.ImpersonateCredentials("impersonated-account@fabled-ray-104117.iam.gserviceaccount.com"))
    ts := creds.TokenSource

    tok, err := ts.Token()
    log.Printf("access_token %v", tok.AccessToken)

}
```

codyoss:
@salrashid123 This package is meant to be a replacement for that option. The TokenSource generated from here, using ADC and options to configure base credentials, can be passed to any client with [option.WithTokenSource](https://pkg.go.dev/google.golang.org/api/option#WithTokenSource).

martinroberts:
Thank you for breaking this out for those of us calling APIs with no client library.  And thanks to @salrashid123 for code and advice in the interim.

I notice that your docs focus on impersonation but you also support (domain-wide)delegation.  In particular, I have been caught out by the need for the targetPrincipal to have "Service Account Token Creator" (SATC) on itself and think others would benefit from including this detail in the docs.

Let me spell that out:  for a pod with workload identity `WI1` (base credentials) impersonating service account `SA2` (target principal) to act on behalf of user `U3` (subject), we need
- `W1` has SATC role on `SA2`
- `SA2` has SATC role on `SA2` (!)
- `SA2` has been granted relevant scopes and rights by the domain admin for `U3`

impersonate/impersonate_test.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+		{
+			name:            "lifetime over max",
+			targetPrincipal: "foo@project-id.iam.gserviceaccount.com",
+			scopes:          []string{"scope"},
+			lifetime:        3601 * time.Second,
+			wantErr:         true,
+		},
+		{
+			name:            "lifetime over max",
```
tbpg, line 47: Update.

impersonate/impersonate_test.go
```diff
@@ -0,0 +1,102 @@
...
+			targetPrincipal: "foo@project-id.iam.gserviceaccount.com",
+			scopes:          []string{"scope"},
+			wantErr:         false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
tbpg, line 55: Same `tt.name` comment. I don't think it's strictly necessary since these aren't run in parallel, though.

impersonate/impersonate_test.go
```diff
@@ -0,0 +1,102 @@
...
+			saTok := "sa-token"
+			client := &http.Client{
+				Transport: RoundTripFn(func(req *http.Request) *http.Response {
+					log.Println(req.URL.Path)
```
tbpg, line 59: Test probably shouldn't log.

impersonate/impersonate_test.go
```diff
@@ -0,0 +1,102 @@
...
+							ExpireTime:  time.Now().Format(time.RFC3339),
+						}
+						b, err := json.Marshal(&resp)
+						if err != nil {
+							t.Fatalf("unable to marshal response: %v", err)
+						}
+						return &http.Response{
+							StatusCode: 200,
+							Body:       ioutil.NopCloser(bytes.NewReader(b)),
+							Header:     make(http.Header),
```
tbpg, line 72: For some reason, `make` is surprising to me. Used to seeing `&http.Header{}`.

impersonate/integration_test.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	rand.Seed(time.Now().UnixNano())
+	baseKeyFile = os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
+	projectID = os.Getenv("GOOGLE_CLOUD_PROJECT")
+	readerKeyFile = os.Getenv("GCLOUD_TESTS_IMPERSONATE_READER_KEY")
+	readerEmail = os.Getenv("GCLOUD_TESTS_IMPERSONATE_READER_EMAIL")
+	writerEmail = os.Getenv("GCLOUD_TESTS_IMPERSONATE_WRITER_EMAIL")
```
tbpg, line 38: Skip if any of these are unset?

impersonate/user.go
```diff
@@ -0,0 +1,176 @@
...
+	"golang.org/x/oauth2"
+)
+
+func user(ctx context.Context, c Config, client *http.Client) (oauth2.TokenSource, error) {
+	// Default to the longest acceptable value of one hour as the token will
+	// be refreshed automatically if not set.
+	lifetime := 3600 * time.Second
+	if c.Lifetime != 0 {
+		lifetime = c.Lifetime
+	}
```
tbpg, line 28: Feels like this should be consolidated before `user` is called. Can add an `autoRefresh` (or similar) argument to control if the token should auto-refresh.

impersonate/user.go
```diff
@@ -0,0 +1,176 @@
...
+	return oauth2.ReuseTokenSource(nil, u), nil
+}
+
+type claimSet struct {
+	Iss   string `json:"iss"`
+	Scope string `json:"scope,omitempty"`
+	Sub   string `json:"sub,omitempty"`
+	Aud   string `json:"aud"`
+	Iat   int64  `json:"iat"`
```
tbpg, line 57: Should this be `IAt`? Or is it conventionally `Iat`?

impersonate/idtoken.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("impersonate: unable to create request: %v", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := i.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("impersonate: unable to generate ID token: %v", err)
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(io.LimitReader(resp.Body, 1<<20))
```
tbpg, line 113: Is there a concern that the body won't be fully read, as required by `http`?

impersonate/user_test.go
```diff
@@ -0,0 +1,115 @@
...
+		},
+		{
+			name:            "lifetime over max",
+			targetPrincipal: "foo@project-id.iam.gserviceaccount.com",
+			scopes:          []string{"scope"},
+			lifetime:        3601 * time.Second,
+			wantErr:         true,
+		},
+		{
+			name:            "lifetime over max",
```
tbpg, line 47: Update.

tbpg (changes requested):

impersonate/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+		Scopes:          []string{"https://www.googleapis.com/auth/cloud-platform"},
+		// Optionally supply delegates.
+		Delegates: []string{"bar@project-id.iam.gserviceaccount.com"},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Pass an impersonated credential to any function that takes a client
+	// options.
```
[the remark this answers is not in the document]
codyoss: Good call, updated.

impersonate/example_test.go
```diff
@@ -0,0 +1,97 @@
...
+		IncludeEmail:    true,
+		// Optionally supply delegates.
+		Delegates: []string{"bar@project-id.iam.gserviceaccount.com"},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Pass an impersonated credential to any function that takes a client
+	// options.
```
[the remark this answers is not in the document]
codyoss: Ack.

impersonate/idtoken.go
```diff
@@ -0,0 +1,130 @@
...
+	its := impersonatedIDTokenSource{
+		client:          client,
+		targetPrincipal: config.TargetPrincipal,
+		audience:        config.Audience,
+		includeEmail:    config.IncludeEmail,
+	}
+	its.delegates = make([]string, len(config.Delegates))
+	for i, v := range config.Delegates {
+		its.delegates[i] = formatIAMServiceAccountName(v)
+	}
```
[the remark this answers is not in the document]
codyoss: Only because the exact size is known. Was just trying to avoid unnecessary allocation I guess. I can update if you think it improves readability.

impersonate/idtoken.go
```diff
@@ -0,0 +1,130 @@
...
+	return oauth2.ReuseTokenSource(nil, its), nil
```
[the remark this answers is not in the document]
codyoss: Correct, with ReuseTokenSource you never need to send in the first token. This will basically make the TokenSource lazily load the token the first time when it needs to and cache it.

impersonate/idtoken.go
codyoss: I don't think so. This is a pattern that is heavily used in the oauth2 package and also by other packages in this repo. Just more of a defensive practice. I believe a Body being fully read just ensures the underlying http connection can be reused. In practice as long as the service being talked to behave appropriately this will never be an issue.

impersonate/idtoken.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	var generateIDTokenResp generateIDTokenResponse
+	if err := json.Unmarshal(body, &generateIDTokenResp); err != nil {
+		return nil, fmt.Errorf("impersonate: unable to parse response: %v", err)
+	}
+	return &oauth2.Token{
+		AccessToken: generateIDTokenResp.Token,
+		// Generated ID tokens are good for one hour.
+		Expiry: now.Add(1 * time.Hour),
```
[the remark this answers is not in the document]
codyoss: Oddly enough the iamcredentials endpoint for creating ID tokens does not support a lifetime parameter today. The tokens is returns are always just good for one hour. Only thing we could add today is a bool option not to refresh the token. But this too could be added at a later date if it is something users would like.

impersonate/idtoken_test.go
```diff
@@ -0,0 +1,82 @@
...
+		{
+			name:            "works",
+			aud:             "http://example.com/",
+			targetPrincipal: "foo@project-id.iam.gserviceaccount.com",
+			wantErr:         false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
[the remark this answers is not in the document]
codyoss: Updated all to capture name.

impersonate/impersonate.go
```diff
@@ -0,0 +1,180 @@
...
+	htransport "google.golang.org/api/transport/http"
+)
+
+var (
+	iamCredentailsEndpoint = "https://iamcredentials.googleapis.com"
+	oauth2Endpoint         = "https://oauth2.googleapis.com"
+)
+
+// Config for generating impersonated credentials.
+type Config struct {
```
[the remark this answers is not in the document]
codyoss: I like it -- I have had some similar thoughts as well. Updated.

impersonate/impersonate.go
```diff
@@ -0,0 +1,180 @@
...
+	// impersonate. Required.
+	TargetPrincipal string
+	// Scopes that the impersonated credential should have. Required.
+	Scopes []string
+	// Delegates are the service account email addresses in a delegation chain.
+	// Each service account must be granted roles/iam.serviceAccountTokenCreator
+	// on the next service account in the chain. Optional.
+	Delegates []string
+	// Lifetime is the amount of time until the impersonated token expires. If
+	// unset the tokens lifetime will be one hour and be automatically
```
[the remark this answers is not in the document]
codyoss: Ack

impersonate/impersonate.go
```diff
@@ -0,0 +1,180 @@
...
+	if config.Lifetime.Seconds() > 3600 {
+		return nil, fmt.Errorf("impersonate: max lifetime is 3600s")
+	}
+	clientOpts := append(defaultClientOptions(), opts...)
+	client, _, err := htransport.NewClient(ctx, clientOpts...)
+	if err != nil {
+		return nil, err
+	}
+	// If a subject is specified a different auth-flow is initiated to
+	// impersonate as the provided subject(user).
```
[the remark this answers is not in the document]
codyoss: Ack.

impersonate/impersonate.go
```diff
@@ -0,0 +1,180 @@
...
+	if config.Subject != "" {
+		return user(ctx, config, client)
+	}
+
+	// Default to the longest acceptable value of one hour as the token will
+	// be refreshed automatically if not set.
+	lifetime := "3600s"
+	if config.Lifetime != 0 {
+		lifetime = fmt.Sprintf("%.fs", config.Lifetime.Seconds())
+	}
```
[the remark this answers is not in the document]
codyoss: The reason it is not is the auth flow in user does not require this formatting of time, instead it is set as a unix timestamp in the JWT before it gets signed.

impersonate/impersonate_test.go
codyoss: Oops, too much copy paste. Good catch!.

impersonate/impersonate_test.go
codyoss: Ack.

impersonate/impersonate_test.go
codyoss: I have no preference here so updated to `http.Header{}`.

impersonate/integration_test.go
```diff
@@ -0,0 +1,145 @@
...
+	os.Exit(m.Run())
+}
+
+func TestTokenSourceIntegration(t *testing.T) {
```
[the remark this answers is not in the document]
codyoss: There will be but still figuring this out. It requires some extra setup so for now I am manual testing this, will get that code uploaded once I know it is for sure working.

impersonate/integration_test.go
codyoss: Ack.

impersonate/user.go
codyoss: Done.

impersonate/user.go
codyoss: I always have seen it as `iat` so I would assume `Iat`.

impersonate/user_test.go
codyoss: Ack.

impersonate/impersonate.go
```diff
@@ -0,0 +1,182 @@
...
+	// Default to the longest acceptable value of one hour as the token will
+	// be refreshed automatically if not set.
+	lifetime := 3600 * time.Second
+	if config.Lifetime != 0 {
+		lifetime = config.Lifetime
+		// Don't auto-refresh token if a lifetime is configured.
+		isStaticToken = true
+	}
+
+	clientOpts := append(defaultClientOptions(), opts...)
```
salrashid123, line 83: just a qq,   if i had a tokensource already, i could pass that through as an `option` here ...meaning, a developer can create their own tokensource, then use `ts, err:= impersonate.CredentialTokenSource(...,....,...option.WithTokenSource(mytokensource))`    ..is that right.

the reason i ask is that'd allow users to acquire a token by whatever means they want

for eample, in a different implementation, i passed in the `RootTokenSource` into the struct that created the impersoanted token:

```golang
targetPrincipal := "impersonated-account@fabled-ray-104117.iam.gserviceaccount.com"
lifetime := 30 * time.Second
delegates := []string{}
targetScopes := []string{"https://www.googleapis.com/auth/devstorage.read_only",
	"https://www.googleapis.com/auth/cloud-platform"}
rootTokenSource, err := google.DefaultTokenSource(ctx,
	"https://www.googleapis.com/auth/iam")
if err != nil {
	log.Fatal(err)
}
tokenSource, err := sal.ImpersonatedTokenSource(
	&sal.ImpersonatedTokenConfig{
		RootTokenSource: rootTokenSource,
		TargetPrincipal: targetPrincipal,
		Lifetime:        lifetime,
		Delegates:       delegates,
		TargetScopes:    targetScopes,
	},
)
if err != nil {
	log.Fatal(err)
}

creds := &google.Credentials{
	TokenSource: tokenSource,
}
```

codyoss:
@salrashid123 

> ust a qq, if i had a tokensource already, i could pass that through as an option here ...meaning, a developer can create their own tokensource, then use ts, err:= impersonate.CredentialTokenSource(...,....,...option.WithTokenSource(mytokensource)) ..is that right.

That is correct. If you used the TokenSource option for that method that would override ACD and be used as the base credential.

impersonate/impersonate.go
codyoss: @salrashid123

ust a qq, if i had a tokensource already, i could pass that through as an option here ...meaning, a developer can create their own tokensource, then use ts, err:= impersonate.CredentialTokenSource(...,....,...option.WithTokenSource(mytokensource)) ..is that right.

That is correct. If you used the TokenSource option for that method that would override ACD and be used as the base credential.

impersonate/integration_test.go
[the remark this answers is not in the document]
tbpg: Sounds good. Fine to do in a future PR as long as we have a tracking issue and do actually implement the test.

tbpg (approved):
LGTM. Would be great to have users try this out before releasing into a non-prerelease version (i.e. `v0.81.0-impersonate-beta` before `v0.81.0` (I forget the exact syntax)) .

impersonate/integration_test.go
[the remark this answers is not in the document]
codyoss: Done.

natalysheinin:
Just tried this and it worked as expected. Thank-you so much for pushing this feature through.

hanneshayashi:
Works like a charm. Thanks for this!

codyoss:
@salrashid123

> never really clear how to use our tokensources with grpc connections
For use with our gRPC based client libraries you can use the option.WithTokenSource. If you are talking about manually dialing what you did in your gist is fine dialing with grpc.WithPerRPCCredentials. You could use the oauth.TokenSource directly instead of using a wrapper if you wanted. We wrap it in our transport code because we add a little bit of extra metadata.

salrashid123:
yeah, the grpc test was using this library to get an idtoken for a non GCP service api (i.e an api that i run that is grpc that happens to accept a google id_token (and in this case, the idtoken is  for an impersonated creds).

but yeah, thanks.  the `oauth.TokenSource{}` also works here.

```golang
	conn, err = grpc.Dial(grpcHost+":443",
		grpc.WithTransportCredentials(ce),
		grpc.WithPerRPCCredentials(oauth.TokenSource{
			TokenSource: gRPCidTokenSource,
		}),
	)
```

upodroid:
I'm implementing the new impersonate package in https://github.com/hashicorp/terraform/pull/28296 and it looks like the package only allows the initial credentials to be an ADC. Terraform supports static accesstoken(static TokenSource) or json loaded with option.WithCredentialsJSON which this package doesn't support.

https://github.com/hashicorp/terraform/blob/b7fb533bd27d63e582489906a63bf1b25ecd5e49/backend/remote-state/gcs/backend.go#L152

codyoss:
@upodroid You should be able to pass in any TokenSource by doing something like: `impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{...}, option.WithTokenSource(anyTokenSource))` This will use the provided TokenSource instead of creds found via ADC for the base credentials in the impersonation flow. Does that solve your issue?

upodroid:
That worked for me, thanks.

upodroid:
btw, v0.44.0 doesn't include the impersonate package. It would be nice to see it in v0.45.0

codyoss:
@upodroid Yes, v0.44.0 got cut sooner than originally planned. I hope to get this in the next main release. I will ping this issue and update the release notes accordingly when that happens.

Issue #950, opened 2021-03-25
docs: improve godoc generation for functions


kortschak:
The godoc comments that are generated for the Go API client code lacks information that is present for other languages, for example python.

Take for example the YouTube live broadcasts API functions. In the `liveBroadCasts` [`insert`](https://googleapis.github.io/google-api-python-client/docs/dyn/youtube_v3.liveBroadcasts.html#insert) documentation all the input parameters are described.

```
insert(part, body=None, onBehalfOfContentOwner=None, onBehalfOfContentOwnerChannel=None, x__xgafv=None)

Inserts a new stream for the authenticated user.

Args:
  part: string, The *part* parameter serves two purposes in this operation. It identifies the properties that the write operation will set as well as the properties that the API response will include. The part properties that you can include in the parameter value are id, snippet, contentDetails, and status. (required) (repeated)
  body: object, The request body.
    The object takes the form of:

<yada...>
```

In the Go client documentation for the cognate [`Insert`](https://pkg.go.dev/google.golang.org/api/youtube/v3#LiveBroadcastsService.Insert) method the input parameters are not described completely.

> ```
> func (r *LiveBroadcastsService) Insert(part []string, livebroadcast *LiveBroadcast) *LiveBroadcastsInsertCall
> ```
>
> Insert: Inserts a new stream for the authenticated user. 

The composite type `LiveBroadcast` documents some of the inputs, but the `parts` parameter is left undocumented, although this information is available to the code generator as seen in the python docs.

Please emit documentation comments for input parameters that do not have documentation already associated with them (i.e. the structs defined in google.golang.org/api/...).

codyoss:
Thanks for the suggestion. I will take a look and see how other languages are handling this today.

PR #964, opened 2021-03-31
docs: include documentation for required parameters 

codyoss:
Currently it is unclear how to use some of the functions that
return calls. Their required parameters are undocumented. This
means a user of the library either needs to go look at the official
docs page for the api or find the corresponding do method in the
code and look for its private discovery json comment. This commit
starts to add parameter descriptions for required fields, those
that are passed to the Call method.

Fixes: #950


codyoss:
@tbpg The first commit contains the actual code changes to the generator. LMK what you think of the formatting. I am open to suggestions!

accessapproval/v1/accessapproval-gen.go
```diff
@@ -960,6 +964,11 @@ type FoldersUpdateAccessApprovalSettingsCall struct {
 // UpdateAccessApprovalSettings: Updates the settings associated with a
 // project, folder, or organization. Settings to update are determined
 // by the value of field_mask.
+//
+// - name: The resource name of the settings. Format is one of: *
+// "projects/{project}/accessApprovalSettings" *
+// "folders/{folder}/accessApprovalSettings" *
+// "organizations/{organization}/accessApprovalSettings"
```
tbpg, line 971: Can we make this show up properly?

abusiveexperiencereport/v1/abusiveexperiencereport-gen.go
```diff
@@ -259,6 +259,9 @@ type SitesGetCall struct {
 }
 
 // Get: Gets a site's Abusive Experience Report summary.
+//
+// - name: The name of the site whose summary to get, e.g.
+// `sites/http%3A%2F%2Fwww.google.com%2F`. Format: `sites/{site}`
```
tbpg, line 264: Annoying question: is there a way to indent the second line?

tbpg, line 264: I wonder if we should add a period at the end if one isn't there?

tbpg (commented):
Looks great!

abusiveexperiencereport/v1/abusiveexperiencereport-gen.go
codyoss: That is a good idea!

tbpg (approved):
LGTM.

kortschak:
Thanks

PR #973, opened 2021-04-07
fix(compute): make MetadataItems.Value a pointer

codyoss:
This is already the case in both the beta and v1 surface.


codyoss:
cc @illfelder

illfelder:
Thank you for making this update - it looks good to me!

tbpg (approved):
LGTM for consistency with other versions -- essentially a bug at this point.

PR #985, opened 2021-04-12
docs: make warning more direct

elharo:
@codyoss


codyoss:
This file is autogenerated based off of the public discovery doc spec. Once changes are rolled out this code will be automatically updated.

elharo:
so where is the public discovery doc spec?

codyoss:
After changes are made internally things will propagate out to the public endpoint: https://dataflow.googleapis.com/$discovery/rest?version=v1b3

elharo:
Can you be specific? Which file where is the source of this? (I wish we didn't check generated code into repos...)

codyoss:
Messaged back directly.

PR #987, opened 2021-04-20
test(byoid): AWS integration test for 3rd party access to Google Client Libraries

gIthuriel:


integration-tests/byoid/setup.sh
```diff
@@ -47,7 +47,7 @@ gcloud beta iam workload-identity-pools providers create-oidc $oidc_provider_id
     --workload-identity-pool=$pool_id \
     --issuer-uri="https://accounts.google.com" \
     --location="global" \
-    --attribute-mapping="google.subject=assertion.sub"
+    --attribute-mapp`ing="google.subject=assertion.sub"
```
bojeil-google, line 50: typo, please fix.

integration-tests/byoid/integration_test.go
```diff
@@ -174,6 +210,7 @@ type credentialSource struct {
 	Headers map[string]string `json:"headers,omitempty"`
 
 	EnvironmentID               string `json:"environment_id,omitempty"`
+	RegionURL                   string `json:"region_url"`
```
ScruffyProdigy, line 213: I don't think you use this

integration-tests/byoid/integration_test.go
```diff
@@ -41,11 +51,32 @@ import (
 const (
 	envCredentials  = "GCLOUD_TESTS_GOLANG_KEY"
 	envAudienceOIDC = "GCLOUD_TESTS_GOLANG_AUDIENCE_OIDC"
+	envAudienceAWS  = "GCLOUD_TESTS_GOLANG_AUDIENCE_AWS"
 	envProject      = "GCLOUD_TESTS_GOLANG_PROJECT_ID"
 )
 
+// The following are specifically for AWS integration tests.
```
ScruffyProdigy, line 58: I don't think you use this

ScruffyProdigy (commented):
LGTM

codyoss:
@gIthuriel got your other PR merged. After you merge your changes I will take a look at this PR.

integration-tests/byoid/integration_test.go
```diff
@@ -37,9 +37,14 @@ import (
 	"log"
 	"net/http"
 	"net/http/httptest"
+	neturl "net/url"
```
codyoss, line 40: nit: Don't rename this import. Rename a variable if required.

integration-tests/byoid/integration_test.go
```diff
@@ -37,9 +37,14 @@ import (
...
 	"os"
+	"regexp"
 	"testing"
 
+	"golang.org/x/oauth2/google"
+
+	"google.golang.org/api/dns/v1"
```
codyoss, line 47: This import is listed twice.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +240,84 @@ func TestURLBasedCredentials(t *testing.T) {
...
 	})
 }
+
+// Tests to make sure AWS based external credentials work properly.
+func TestAWSBasedCredentials(t *testing.T) {
+	data := neturl.Values{}
+	data.Set("audience", clientID)
+	data.Set("includeEmail", "true")
+
+	client, err := google.DefaultClient(context.Background(), "https://www.googleapis.com/auth/cloud-platform")
```
codyoss, line 250: Please check the error.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +240,84 @@ func TestURLBasedCredentials(t *testing.T) {
...
+	resp, err := client.PostForm(fmt.Sprintf("https://iamcredentials.googleapis.com/v1/%s:generateIdToken", clientID), data)
```
codyoss, line 251: Please check the error.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +240,84 @@ func TestURLBasedCredentials(t *testing.T) {
...
+	if resp.StatusCode != 200 {
+		t.Errorf("byoid: failed to get Google ID token for AWS test: %v", err)
+	}
+
+	var res map[string]interface{}
+
+	json.NewDecoder(resp.Body).Decode(&res)
```
codyoss, line 258: Please check the error.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +240,84 @@ func TestURLBasedCredentials(t *testing.T) {
...
+
+	data = neturl.Values{}
+	data.Set("Action", "AssumeRoleWithWebIdentity")
+	data.Set("Version", "2011-06-15")
+	data.Set("DurationSeconds", "3600")
+	data.Set("RoleSessionName", os.Getenv("GCLOUD_TESTS_GOLANG_AWS_ROLE_NAME"))
+	data.Set("RoleArn", os.Getenv("GCLOUD_TESTS_GOLANG_AWS_ROLE_ID"))
+	data.Set("WebIdentityToken", token.(string))
+
+	resp, err = http.PostForm("https://sts.amazonaws.com/", data)
```
codyoss, line 273: Please check the error.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +240,84 @@ func TestURLBasedCredentials(t *testing.T) {
...
+// Tests to make sure AWS based external credentials work properly.
+func TestAWSBasedCredentials(t *testing.T) {
+	data := neturl.Values{}
+	data.Set("audience", clientID)
+	data.Set("includeEmail", "true")
+
+	client, err := google.DefaultClient(context.Background(), "https://www.googleapis.com/auth/cloud-platform")
+	resp, err := client.PostForm(fmt.Sprintf("https://iamcredentials.googleapis.com/v1/%s:generateIdToken", clientID), data)
+	if resp.StatusCode != 200 {
+		t.Errorf("byoid: failed to get Google ID token for AWS test: %v", err)
```
codyoss, line 253: This prefix convention is only used in non-test code when returning errors. Please see this example for good testing errors: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +240,84 @@ func TestURLBasedCredentials(t *testing.T) {
...
+	resp, err = http.PostForm("https://sts.amazonaws.com/", data)
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		t.Errorf("byoid: failed to parse response body from AWS: %v", err)
+	}
+
+	bodyString := string(bodyBytes)
+
+	// Using regex instead of loading the full XML response because I only need a few fields.
+	SessTok := regexp.MustCompile("<SessionToken>(.*)</SessionToken>")
```
codyoss, line 282: You should not need to load the full payload. Just define a minimal struct with xml struct tags then use the xml package to parse.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +240,84 @@ func TestURLBasedCredentials(t *testing.T) {
...
+		Audience:                       awsAudience,
+		SubjectTokenType:               "urn:ietf:params:aws:token-type:aws4_request",
+		TokenURL:                       "https://sts.googleapis.com/v1/token",
+		ServiceAccountImpersonationURL: fmt.Sprintf("https://iamcredentials.googleapis.com/v1/%s:generateAccessToken", clientID),
+		CredentialSource: credentialSource{
+			EnvironmentID:               "aws1",
+			RegionalCredVerificationURL: "https://sts.us-east-1.amazonaws.com?Action=GetCallerIdentity&Version=2011-06-15",
+		},
+	})
+}
```
codyoss, line 323: nit: add newline.

codyoss (changes requested):

integration-tests/byoid/integration_test.go
gIthuriel: As a head's up, this rename is actually copied from the original implementation in `oauth2`.  Should I rename it there as well?

integration-tests/byoid/integration_test.go
gIthuriel: That's bizarre, it isn't listed twice locally.  Let me try and fix that.

integration-tests/byoid/integration_test.go
gIthuriel: +
if you wish, I can go learn the XML package and use that instead.  I thought this was avoiding loading the full payload, honesty, but maybe I misinterpreted what that meant.  I thought that I'd have to define an XML struct with tags supporting everything in the response in order to do this, which would be quite a lot.  I tried to model  this off of the Python implementation.

integration-tests/byoid/integration_test.go
gIthuriel: Do you mean a newline at the end of the file?  I already have that locally.

integration-tests/byoid/integration_test.go
```diff
@@ -202,3 +216,115 @@ func TestFileBasedCredentials(t *testing.T) {
...
+
+// Tests to make sure AWS based external credentials work properly.
+func TestAWSBasedCredentials(t *testing.T) {
+	data := url.Values{}
+	data.Set("audience", clientID)
+	data.Set("includeEmail", "true")
+
+	client, err := google.DefaultClient(context.Background(), "https://www.googleapis.com/auth/cloud-platform")
+	if err != nil {
+		t.Errof("Failed to create default client: %v", err)
```
codyoss, line 250: Would you step through and reevaulate some of these t.Errors, I think some should be t.Fatal such as this one. If there is an error creating the client the rest of the test can't run.

integration-tests/byoid/integration_test.go
codyoss: I think the test would be cleaner if the xml package was used. The package has some great examples, such as this one: https://pkg.go.dev/encoding/xml#example-Unmarshal. It should have the same feel as the json package. You only need to define the fields that you wish to parse.

integration-tests/byoid/integration_test.go
```diff
@@ -30,9 +35,15 @@ import (
...
 	"log"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"os"
+	"regexp"
 	"testing"
 
+	"golang.org/x/oauth2/google"
+
```
codyoss, line 46: nit: remove space.

codyoss (changes requested):

integration-tests/byoid/integration_test.go
codyoss: Yes, it must not have been pushed.

integration-tests/byoid/integration_test.go
codyoss: We can fix it next time there are some changes to be made over there.

integration-tests/byoid/integration_test.go
gIthuriel: Doesn't t.Errorf automatically halt the execution of the test and return an error?

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +236,104 @@ func TestURLBasedCredentials(t *testing.T) {
...
+	resp, err = http.PostForm("https://sts.amazonaws.com/", data)
+	if err != nil {
+		t.Fatalf("Failed to post data to AWS: %v", err)
+	}
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatalf("Failed to parse response body from AWS: %v", err)
+	}
+
+	type Credentials struct {
```
codyoss, line 286: If I am understanding what your xml should look like, this can be simplified a bit to something like this: https://play.golang.org/p/hz6hUSDKBfX

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +236,104 @@ func TestURLBasedCredentials(t *testing.T) {
...
+	}
+	type AssumeRoleWithWebIdentityResponse struct {
+		XMLName                         xml.Name                        `xml:"AssumeRoleWithWebIdentityResponse"`
+		AssumeRoleWithWebIdentityResult AssumeRoleWithWebIdentityResult `xml:"AssumeRoleWithWebIdentityResult"`
+	}
+
+	respFields := AssumeRoleWithWebIdentityResponse{}
+	err = xml.Unmarshal(bodyBytes, &respFields)
+
+	if respFields.AssumeRoleWithWebIdentityResult.Credentials.SessionToken != "" {
```
codyoss, line 302: Can this be refactored to check all variables before setting environment vars. That was the happy path is not indented in.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +236,104 @@ func TestURLBasedCredentials(t *testing.T) {
...
+	type AssumeRoleWithWebIdentityResult struct {
+		Credentials Credentials `xml:"Credentials"`
+	}
+	type AssumeRoleWithWebIdentityResponse struct {
+		XMLName                         xml.Name                        `xml:"AssumeRoleWithWebIdentityResponse"`
+		AssumeRoleWithWebIdentityResult AssumeRoleWithWebIdentityResult `xml:"AssumeRoleWithWebIdentityResult"`
+	}
+
+	respFields := AssumeRoleWithWebIdentityResponse{}
+	err = xml.Unmarshal(bodyBytes, &respFields)
```
codyoss, line 300: Please check the error.

codyoss (changes requested):

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +236,104 @@ func TestURLBasedCredentials(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to generate an ID token: %v", err)
+	}
+	if resp.StatusCode != 200 {
+		t.Fatalf("Failed to get Google ID token for AWS test: %v", err)
+	}
+
+	var res map[string]interface{}
+
+	err = json.NewDecoder(resp.Body).Decode(&res)
```
codyoss, line 260: nit:  this can all be one expression.

```go
if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
   ...
}
```

integration-tests/byoid/integration_test.go
gIthuriel: I figured you'd want a cleaner separated implementation, but no worries- I just changed it.

integration-tests/byoid/integration_test.go
```diff
@@ -231,3 +236,90 @@ func TestURLBasedCredentials(t *testing.T) {
...
+		t.Fatalf("Failed to parse response body from AWS: %v", err)
+	}
+
+	var respVars struct {
+		SessionToken    string `xml:"AssumeRoleWithWebIdentityResult>Credentials>SessionToken"`
+		SecretAccessKey string `xml:"AssumeRoleWithWebIdentityResult>Credentials>SecretAccessKey"`
+		AccessKeyID     string `xml:"AssumeRoleWithWebIdentityResult>Credentials>AccessKeyId"`
+	}
+
+	err = xml.Unmarshal(bodyBytes, &respVars)
```
codyoss, line 291: Pleas check error, similar to json usage aboce.

codyoss (approved):

PR #1028, opened 2021-05-18
fix(transport): increase ReadIdleTimeout

tritone:
Increase this value to comply with CFE rate limit on pings.

Closes googleapis/google-cloud-go#4121


shollyman (approved):
Disclaimer on LGTM: I don't claim to have full awareness of the magic here, but that seems like a more sensible default generally.

codyoss (approved):
LGTM

PR #1038, opened 2021-05-24
fix(transport): Perform mtls endpoint switching with custom http client

shinfan:
The terraform CLI requires this change in order to support mTLS.

For existing users that provide custom http clients today, this change is invisible since it is gated by the mTLS environment variable.


codyoss (approved):
LGTM

PR #1040, opened 2021-05-25
fix(transport): add ShouldUseMTLSEndpoint(...)

shinfan:
Create ShouldUseMTLSEndpoint(...) which returns if the MTLS endpoint should be used. This method is needed for Terraform to onboard mTLS and we do not want to Terraform to write its own logic to detect which endpoint to use.

### DIFF
--- a/transport/cert/default_cert.go
+++ b/transport/cert/default_cert.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 )
@@ -30,6 +31,10 @@ import (
 const (
 	metadataPath = ".secureConnect"
 	metadataFile = "context_aware_metadata.json"
+
+	mTLSModeAlways = "always"
+	mTLSModeNever  = "never"
+	mTLSModeAuto   = "auto"
 )
 
 // defaultCertData holds all the variables pertaining to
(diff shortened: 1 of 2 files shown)


transport/cert/dca.go
```diff
@@ -43,7 +42,7 @@ const (
 // GetClientCertificateSourceAndEndpoint is a convenience function that invokes
 // getClientCertificateSource and getEndpoint sequentially and returns the client
 // cert source and endpoint as a tuple.
-func GetClientCertificateSourceAndEndpoint(settings *internal.DialSettings) (cert.Source, string, error) {
+func GetClientCertificateSourceAndEndpoint(settings *internal.DialSettings) (Source, string, error) {
```
codyoss, line 45: I don't think this package can be made public as is. The input to this function is a struct that can't be made public. I think for now a little bit of copying is the right answer. This change would add quite a bit of public surface.

codyoss (changes requested):

shinfan:
Thanks for review. PTAL. @codyoss 

Not related to this PR but I am a little concerned about the package name as I feel we should have the top-level package called "mtls" instead of "cert" to be more future proof, since we will add some more mtls related features later. Is this something too late to change?

codyoss:
What is the use case where these methods need to be exposed? We already allow for returning a `http.Client` that uses the certSource to setup the TLSConfig. I guess I am lacking the proper context here, is there a design doc that can be shared with me? Are other languages exposing similar features?

codyoss:
Closing this PR for now as we decided to take a different direction.

PR #1065, opened 2021-06-10
fix: URL's in `ProjectsGetRemoteConfigCall`/`ProjectsUpdateRemoteConfigCall`

sphinkx:
According to [discovery](https://firebaseremoteconfig.googleapis.com/$discovery/rest?version=v1) calls for `getRemoteConfig`/`updateRemoteConfig` should make to URL `v1/projects/{projectsId}/remoteConfig`, not `v1/{projectsId}/remoteConfig` (only broken robot has seen on this adress).


codyoss:
Hey @sphinkx thanks for the report. This client is auto-generated so manual changes can't be made to it. It appears that even though this client does have a public discovery document it is no longer found in the [index of apis](https://www.googleapis.com/discovery/v1/apis) we generate. I have reached out to internal teams to see if this is intentional. Looking at the revision of this API it appears it has not been updated in over 3 years in this repo.

I will report back when I hear more.

codyoss:
Closing for reason listed above.

PR #1075, opened 2021-06-16
feat(internal): [AIP-4111] support scopes for self-signed JWT auth flow

shinfan:
- A **self-signed JWT flow** will be executed if the following conditions are met:
  - One of the following is true:
    (a) The scope for self-signed JWT flow is enabled (`EnableJwtWithScope` == true)
    (b) Custom audiences are explicitly provided by users
    (c) No scopes are provided at all
  - No service account impersonation
- Otherwise, executes **standard OAuth 2.0 flow** as fallback
- Following cases will result in error: 
   - Neither scopes nor audiences are available with the service account JSON
   - Malformed or empty JSON input

More information for self-signed JWT at: https://google.aip.dev/auth/4111


internal/settings.go
```diff
@@ -24,6 +24,7 @@ type DialSettings struct {
 	DefaultMTLSEndpoint string
 	Scopes              []string
 	DefaultScopes       []string
+	EnableScopeForJWT   bool
```
andyrzhao, line 27: "UseJWTAccessWithScope" based on our discussion for AIP 4111

internal/creds.go
```diff
@@ -102,16 +97,35 @@ func credentialsFromJSON(ctx context.Context, data []byte, ds *DialSettings) (*g
 	return cred, err
```
andyrzhao, line 97: The code flow is a bit inconsistent with the function comment. Function comment implies there are only two cases, OAuth and SelfSignedJWT, yet in code there is currently a 3rd case where neither OAuth or JWTFlow is true, and the cred is straight-up returned, which is the same behavior as when OAuth is true. Should we re-arrange the code flow to make isJWTFlow the first "if", and the rest can straight-up fall back to OAuth? (In other words, isSelfSignedJWTFlow can be updated with the logic in isOAuthFlow and isOAuthFlow function can be removed.)

andyrzhao (changes requested):

internal/creds.go
```diff
@@ -64,54 +65,63 @@ const (
...
 	cred, err := google.CredentialsFromJSON(ctx, data, ds.GetScopes()...)
 	if err != nil {
 		return nil, err
 	}
-	// Standard OAuth 2.0 Flow
-	if len(data) == 0 ||
-		len(ds.Scopes) > 0 ||
-		(ds.DefaultAudience == "" && len(ds.Audiences) == 0) ||
-		ds.ImpersonationConfig != nil ||
-		ds.Endpoint != "" {
```
codyoss, line 83: Is this meant to go away? Endpoint now can work with self-signed /w scopes but it does not when aud is used.

internal/creds.go
```diff
@@ -64,54 +65,63 @@ const (
...
+		if err := json.Unmarshal(data, &f); err != nil {
+			return false, err
+		}
+		return f.Type == serviceAccountKey, nil
+	}
+	return false, nil
+}
+
+func selfSignedJWTTokenSource(data []byte, audience string, scopes []string) (oauth2.TokenSource, error) {
+	if len(scopes) > 0 {
```
codyoss, line 116: In the case of client libraries this will always be true since we always pass default scopes for all clients.

I believe this differs from how the design doc is laid out. If this is what should be done the doc should be updated. The doc states: scope -> aud -> default scope.

codyoss (changes requested):
Thanks for the PR, a few suggestions and comments.

internal/creds.go
```diff
@@ -64,54 +65,63 @@ const (
...
+// - Otherwise, executes standard OAuth 2.0 flow
+// More details: google.aip.dev/auth/4111
 func credentialsFromJSON(ctx context.Context, data []byte, ds *DialSettings) (*google.Credentials, error) {
+	// By default, a standard OAuth 2.0 token source is created
 	cred, err := google.CredentialsFromJSON(ctx, data, ds.GetScopes()...)
 	if err != nil {
 		return nil, err
 	}
-	// Standard OAuth 2.0 Flow
-	if len(data) == 0 ||
```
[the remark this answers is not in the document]
shinfan: This nil check is unnecessary here since this method should be only called when there is actually a JSON. And L78 `google.CredentialsFromJSON` only returns nil with an error.

internal/creds.go
```diff
@@ -64,54 +65,63 @@ const (
...
 }
 
-func selfSignedJWTTokenSource(data []byte, defaultAudience string, audiences []string) (oauth2.TokenSource, error) {
-	audience := defaultAudience
-	if len(audiences) > 0 {
-		// TODO(shinfan): Update golang oauth to support multiple audiences.
-		if len(audiences) > 1 {
-			return nil, fmt.Errorf("multiple audiences support is not implemented")
+func isSelfSignedJWTFlow(data []byte, ds *DialSettings) (bool, error) {
+	if (ds.UseJwtWithScope || len(ds.Audiences) > 0) &&
```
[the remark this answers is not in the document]
shinfan: We don't need to check the default aud here because it does not affect the decision of which flow to use. The self-signed JWT flow is used when:

1) The scope for JWT is enabled
or
2) The audience is provided by the user (not the default audience, which is provided by clients).

internal/creds.go
shinfan: Right I don't think we need this anymore. When we use self-signed JWT there are two conditions:

1) Scope is enabled for self-signed JWT: in this case the scope will be used so custom endpoint will not an issue.
2) User provided audience: the endpoint is not an issue either since the user is responsible of the audience.

internal/creds.go
shinfan: Good point. I think the correct order should be:

custom aud or scope -> default scope -> default aud

I will make this clear in the doc.

option/internaloption/internaloption.go
```diff
@@ -94,3 +94,15 @@ func (w withDefaultScopes) Apply(o *internal.DialSettings) {
 	o.DefaultScopes = make([]string, len(w))
 	copy(o.DefaultScopes, w)
 }
+
+// UseJwtWithScope returns a ClientOption that specifies if scope can be used
+// with self-signed JWT.
+func UseJwtWithScope(useScope bool) option.ClientOption {
```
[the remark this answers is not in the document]
shinfan: Renamed to EnableJwtWithScope. What do you mean bool is not needed here? The other similar option `EnableDirectPath` is also a bool.

internal/creds.go
```diff
@@ -64,54 +65,64 @@ const (
...
 }
 
-func selfSignedJWTTokenSource(data []byte, defaultAudience string, audiences []string) (oauth2.TokenSource, error) {
-	audience := defaultAudience
-	if len(audiences) > 0 {
-		// TODO(shinfan): Update golang oauth to support multiple audiences.
-		if len(audiences) > 1 {
-			return nil, fmt.Errorf("multiple audiences support is not implemented")
+func isSelfSignedJWTFlow(data []byte, ds *DialSettings) (bool, error) {
+	if (ds.EnableJwtWithScope || ds.HasCustomAudience()) &&
```
andyrzhao, line 100: Hmm, what if EnableJwtWithScope is false and scopes is empty and no custom audience is specified by user? Shouldn't we still try SelfSignedJWTFlow with the default audience?

internal/creds.go
```diff
@@ -64,54 +65,64 @@ const (
...
+	}
+	return false, nil
+}
+
+func selfSignedJWTTokenSource(data []byte, ds *DialSettings) (oauth2.TokenSource, error) {
+	if len(ds.GetScopes()) > 0 && !ds.HasCustomAudience() {
+		// Scopes are preferred in self-signed JWT unless the scope is not available
+		// or a custom audience is used.
+		return google.JWTAccessTokenSourceWithScope(data, ds.GetScopes()...)
+	} else if ds.GetAudience() != "" {
```
andyrzhao, line 120: GetAudience will either return custom audience or the default audience so this IF will always be true right? I guess I'm just not familiar with the concept of default audience - is default audience expected to be empty sometimes? Eitherway, I think the test cases needs to be expanded to cover both custom audience and default audience scenarios.

shinfan: Yea this should be true otherwise we will return error in the final else block.

internal/creds.go
shinfan: Good point. I don't think in practice the scope will ever be empty since we always provide default scopes in our clients, but it does not harm to do the extra check.

shinfan:
Thanks for review. @codyoss @andyrzhao PTAL

internal/creds.go
```diff
@@ -62,56 +63,68 @@ const (
...
-// - A self-signed JWT auth flow will be executed if: the data file is a service
-//   account, no user are scopes provided, an audience is provided, a user
-//   specified endpoint is not provided, and credentials will not be
-//   impersonated.
+// - A self-signed JWT flow will be executed if the following conditions are
+// met:
+//   (1) At least one of the following is true:
+//       (a) Scope for self-signed JWT flow is enabled
+//       (b) Audiences are explicitly provided by users
+//       (b) No scope is provided
```
andyrzhao, line 73: should be (c). Btw, I think it would read clearer if the ordering is "a) no scope provided b) JWT with scope is enabled c) audiences explicitly provided.

internal/creds.go
andyrzhao: Ok, I'm just observing that since default audience will be available most of the time, we will rarely reach the "else" part of this block. Maybe updating the final error message to "neither scopes or audience are **available** for the self-signed JWT" is more accurate, as it will capture the edge case where neither custom audience or default audience are available.

andyrzhao (approved):
Looks good overall. Some minor comments.

internal/creds.go
codyoss: Ack

option/internaloption/internaloption.go
```diff
@@ -94,3 +94,15 @@ func (w withDefaultScopes) Apply(o *internal.DialSettings) {
 	o.DefaultScopes = make([]string, len(w))
 	copy(o.DefaultScopes, w)
 }
+
+// EnableJwtWithScope returns a ClientOption that specifies if scope can be used
+// with self-signed JWT.
+func EnableJwtWithScope(enableScope bool) option.ClientOption {
```
codyoss, line 100: What I meant by a previous comment was that this does not need to take a bool. If passed it can be assumed true.

option/internaloption/internaloption.go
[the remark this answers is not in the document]
codyoss: New comment below.

codyoss (approved):
LGTM, thank you for your work on this feature!

Issue #1092, opened 2021-06-28
idtoken package has been broken since v0.49.0 when service_account credential is used


apstndb:
#### Environment details

  - Programming language: Go 
  - OS: macOS
  - Language runtime version: Go 1.16.2
  - Package version: v0.49.0

#### Steps to reproduce

Minimal reproducible code.

```
package main

import (
	"context"
	"log"

	"google.golang.org/api/idtoken"
)

func main() {
	_, err := idtoken.NewTokenSource(context.Background(), "https://example.com")
	if err != nil {
		log.Fatalln(err)
	}
}
```
In `v0.49.0`:
```
$ GOOGLE_APPLICATION_CREDENTIALS=sa.json go run ./
2021/06/28 18:28:33 neither scopes or audience are available for the self-signed JWT
```

In `v0.48.0`, it succeed.

It seems to be related with this comment https://github.com/googleapis/google-api-go-client/pull/1075#discussion_r653919814 and `idtoken` package doesn't specify `WithAudience`.

bweston92:
I opened an issue on the docs that is a temporary work around for this: https://github.com/GoogleCloudPlatform/golang-samples/issues/2125

codyoss:
Thank you for the report, I will investigate why this is happening.

codyoss:
https://github.com/googleapis/google-api-go-client/releases/tag/v0.50.0 includes the fix!

bweston92:
Hey @codyoss thanks for taking the time to put a fix in :) appreciated.

codyoss:
Of course! 😸

PR #1096, opened 2021-06-29
fix(internal): fix self-signed detection with scopes

codyoss:
Removed the requirement on length of scopes passed. Checking the EnableJwtWithScope internal option should be sufficient since it will only be passed with our generated clients that will always include at least the platform scope and a default audience. 
Fixes: #1092

### DIFF
--- a/internal/creds.go
+++ b/internal/creds.go
@@ -99,7 +99,7 @@ func credentialsFromJSON(ctx context.Context, data []byte, ds *DialSettings) (*g
 }
 
 func isSelfSignedJWTFlow(data []byte, ds *DialSettings) (bool, error) {
-	if ((ds.EnableJwtWithScope && len(ds.GetScopes()) > 0) || ds.HasCustomAudience()) &&
+	if (ds.EnableJwtWithScope || ds.HasCustomAudience()) &&
 		ds.ImpersonationConfig == nil {
 		// Check if JSON is a service account and if so create a self-signed JWT.
 		var f struct {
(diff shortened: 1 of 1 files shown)


codyoss:
cc @shinfan

internal/creds.go
```diff
@@ -99,7 +99,7 @@ func credentialsFromJSON(ctx context.Context, data []byte, ds *DialSettings) (*g
 }
 
 func isSelfSignedJWTFlow(data []byte, ds *DialSettings) (bool, error) {
-	if (ds.EnableJwtWithScope || ds.HasCustomAudience() || len(ds.GetScopes()) == 0) &&
+	if ((ds.EnableJwtWithScope && len(ds.GetScopes()) > 0) || ds.HasCustomAudience()) &&
 		ds.ImpersonationConfig == nil {
```
tbpg, line 103: Refactor these two lines to get more descriptive names and avoid lots of logical nesting?

internal/creds.go
```diff
@@ -99,7 +99,7 @@ func credentialsFromJSON(ctx context.Context, data []byte, ds *DialSettings) (*g
 }
 
 func isSelfSignedJWTFlow(data []byte, ds *DialSettings) (bool, error) {
-	if (ds.EnableJwtWithScope || ds.HasCustomAudience() || len(ds.GetScopes()) == 0) &&
+	if ((ds.EnableJwtWithScope && len(ds.GetScopes()) > 0) || ds.HasCustomAudience()) &&
```
shinfan, line 102: I don't feel we need to check for the scope here when scope is enabled for JWT

1) If scope is non-empty, JWT is used
2) If scope is empty, audience must be non-empty, in this case JWT is used too.

Does this make sense?

codyoss: I agree, will remove as that is another way to fix this.

internal/creds.go
codyoss: Done

tbpg (approved):

PR #1117, opened 2021-07-17
improved link for ChromeosdevicesListCall.Query()

discentem:
The current [link](https://support.google.com/chrome/a/answer/1698333) doesn't actually seem to provide examples of queries but https://developers.google.com/admin-sdk/directory/v1/guides/manage-chrome-devices does.

### DIFF
--- a/admin/directory/v1/admin-api.json
+++ b/admin/directory/v1/admin-api.json
@@ -450,7 +450,7 @@
               "type": "string"
             },
             "query": {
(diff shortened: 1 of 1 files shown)


codyoss:
Hey @discentem thanks for the suggestion. Unfortunately these files are auto-generated and not the source of truth. I would suggest filing a bug for the product team so they could consider updating the source of truth for these docs. Here is a [link to their issue tracker](https://issuetracker.google.com/issues/new?component=191635&template=823904).

Issue #550, opened 2020-06-23
healthcare: duplicate query param key


nhue2310:
- I was try to search FHIR resource with `_include` query param like below
```
queries := []googleapi.CallOption{}

queryPractitioner := helper.QueryParamOpt{
	Key:   "_include",
	Value: practitionerID,
}
queries = append(queries, queryPractitioner)
queryOrganization := helper.QueryParamOpt{
	Key:   "_include",
	Value: "organizationID",
}
queries = append(queries, queryOrganization)
hs, err := newHealthService()
r, err := hs.Projects.Locations.Datasets.FhirStores.Fhir.Search(parent).Do(queries)
```
### Expect
`https://healthcare.googleapis.com/v1/projects/x/locations/x/datasets/x/fhirStores/x/fhir/PractitionerRole?_include=Practitioner&_include=Organization`
### Real result
`https://healthcare.googleapis.com/v1/projects/x/locations/x/datasets/x/fhirStores/x/fhir/PractitionerRole?_include=Organization`
 => Search result is incorrect
Seem like this libs didn't support it

As you can see on [this line](https://github.com/googleapis/google-api-go-client/blob/8809fe2216f75c184cf682992f90bd98f325c278/internal/gensupport/params.go#L28), lib will replace query with same key.

So, How can i add new duplicate key without replace old one ?

codyoss:
Can you please show how you are using `queryPractitioner` & `queryOrganization` in the snippet provided above.

nhue2310:
@codyoss 
Sorry, i updated detail
`queryPractitioner & queryOrganization` has been appended into queries.

codyoss:
Thanks, I will need to do some digging and then I will get back to you.

tbpg:
cc @noerog

hue-n-gnk:
@codyoss  Hi, how is it going?
It's more than 1 year for the digging. Do you have a plan for fixing it?

codyoss:
@hue-n-gnk Sorry, this slipped off my radar. I have just opened a PR that should unblock this use-case. Thanks for pinging the issue!

PR #1123, opened 2021-07-20
feat(googleapi): support setting arbitrary query parameters

codyoss:
Fixes: #550


tbpg (commented):
I think this had come up before with @broady, so will leave review to him.

tbpg (commented):
Found https://github.com/GoogleCloudPlatform/golang-samples/pull/1015 for some history.

codyoss:
Yeah I am not sure if there is any other API besides healthcare that needs this but it is at least useful/needed for that API I believe. With how CallOptions currently work this is no way for a user to provide a key with multiple values.

googleapi/googleapi.go
```diff
@@ -417,4 +417,23 @@ type traceTok string
...
+	key    string
+	values []string
+}
+
+// QueryParameter allows setting the value(s) of an arbitrary key.
+func QueryParameter(key string, values ...string) CallOption {
+	return queryParameter{key: key, values: append([]string{}, values...)}
+}
+
+// Get will never actually be called -- GetMulti will.
```
broady, line 430: This is a public package, I think this needs more documentation.

maybe add the type to googleapi:

```
type MultiCallOption interface {
   CallOption
   GetMulti() (string, []string)
}
```

then QueryParameter can return a MultiCallOption (or a better name, not sure)
and then document on MultiCallOption the difference in behavior.

codyoss: Done.

broady:
LGTM (not sure why I can't leave an "Approval")

broady (approved):

PR #1124, opened 2021-07-20
test(downscope): add integration tests for token downscoping with Credential Access Boundaries

Galadros:
Testing the code present in `oauth2/google/downscope`.  I also caught a typo in the comments of a previous push- given that I only needed to add a single space, I bundled that change into this PR.


integration-tests/downscope/downscope_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"testing"
+	"time"
+
+	"google.golang.org/api/option"
+
+	"cloud.google.com/go/storage"
```
codyoss, line 19: This module does not currently rely on this module and I think it is best to keep it that way. Can you refactor this test to use `https://pkg.go.dev/google.golang.org/api/storage/v1`. This is what this client uses under the hood.

codyoss (changes requested):
I'll take a closer look after code has been refactored.

integration-tests/downscope/downscope_test.go
```diff
@@ -0,0 +1,150 @@
...
+		})
+	}
+}
+
+// I'm not sure what I should name this according to convention.
+func helper(tt downscopeTest) error {
+	ctx := context.Background()
+
+	// Initializes an accessBoundary
+	AccessBoundaryRules := make([]downscope.AccessBoundaryRule, 0)
```
codyoss, line 124: nit: slices don't need to be initialized. `var AccessBoundaryRules []downscope.AccessBoundaryRule`

integration-tests/downscope/downscope_test.go
```diff
@@ -0,0 +1,150 @@
...
+	storageService, err := storage.NewService(ctx, option.WithTokenSource(downscopedTokenSource))
+	if err != nil {
+		return fmt.Errorf("failed to create the storage service: %v", err)
+	}
+	obj, err := storageService.Objects.Get(bucket1, tt.objectName).Download()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve object from GCP project with error: %v", err)
+	}
+	dat, err := ioutil.ReadAll(obj.Body)
+	defer obj.Body.Close()
```
codyoss, line 144: This should happen right after the error check.

integration-tests/downscope/setup.sh
```diff
@@ -0,0 +1,76 @@
...
+gsutil cp ${first_object} gs://${bucket_id}
+gsutil cp ${second_object} gs://${bucket_id}
+
+echo "Bucket ID: "${bucket_id}
+echo "First object ID: "${first_object}
+echo "Second object ID: "${second_object}
+
+# Cleanup
+rm ${first_object}
+rm ${second_object}
```
codyoss, line 76: nit: add newline.

integration-tests/downscope/downscope_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+const (
+	envServiceAccountFile = "GCLOUD_TESTS_GOLANG_KEY"
+	rootTokenScope        = "https://www.googleapis.com/auth/cloud-platform"
+)
+
+var (
+	object1 = "cab-first-c45wknuy.txt"
+	object2 = "cab-second-c45wknuy.txt"
+	bucket1 = "dulcet-port-762"
```
codyoss, line 33: nit: Just `bucket`

integration-tests/downscope/downscope_test.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	for _, tt := range downscopeTests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := helper(tt)
+			// If a test isn't supposed to fail, it shouldn't fail.
+			if !tt.expectError && err != nil {
+				t.Errorf("test case %v should have succeeded, but instead returned %v", tt.name, err)
+			} else if tt.expectError && err == nil { // If a test is supposed to fail, it should return a non-nil error.
+				t.Errorf(" test case %v should have returned an error, but instaed returned nil", tt.name)
```
codyoss, line 113: nit: typo `instead`

integration-tests/downscope/downscope_test.go
```diff
@@ -0,0 +1,150 @@
...
+	storage "google.golang.org/api/storage/v1"
+)
+
+const (
+	envServiceAccountFile = "GCLOUD_TESTS_GOLANG_KEY"
+	rootTokenScope        = "https://www.googleapis.com/auth/cloud-platform"
+)
+
+var (
+	object1 = "cab-first-c45wknuy.txt"
```
codyoss, line 31: These could be const as well.

integration-tests/downscope/downscope_test.go
```diff
@@ -0,0 +1,150 @@
...
+		// This line runs all of our individual tests
+		os.Exit(m.Run())
+	}
+	ctx := context.Background()
+	credentialFileName := os.Getenv(envServiceAccountFile)
+	credentialsFileData, err := os.ReadFile(credentialFileName)
+	if err != nil {
+		log.Fatalf("failed to open credentials file: %v", err)
+	}
+	rootCredential, err = google.CredentialsFromJSON(ctx, credentialsFileData, rootTokenScope)
```
codyoss, line 51: You could use https://pkg.go.dev/google.golang.org/api/transport#Creds with the scope and creds file option to simplify this a bit.

codyoss (changes requested):
Thanks, just a couple of things.

integration-tests/downscope/downscope_test.go
Galadros: Thanks, just changed it.  A few of the comments you've made are because the structure of the code changed a few times and I forgot to update some legacy decisions that were made.

integration-tests/downscope/downscope_test.go
```diff
@@ -105,23 +109,24 @@ func TestDownscopedToken(t *testing.T) {
...
-				t.Errorf(" test case %v should have returned an error, but instaed returned nil", tt.name)
+				t.Errorf(" test case %v should have returned an error, but instead returned nil", tt.name)
 			}
 		})
 	}
 }
 
 // I'm not sure what I should name this according to convention.
-func helper(tt downscopeTest) error {
+func helper(t *testing.T, tt downscopeTest) error {
```
codyoss, line 124: nit: This name should still needs to be updated to describe what this function is doing.

integration-tests/downscope/downscope_test.go
```diff
@@ -136,15 +141,14 @@ func helper(tt downscopeTest) error {
 	if err != nil {
 		return fmt.Errorf("failed to create the storage service: %v", err)
 	}
-	obj, err := storageService.Objects.Get(bucket1, tt.objectName).Download()
+	obj, err := storageService.Objects.Get(bucket, tt.objectName).Download()
```
codyoss, line 144: This returns a response I think, not an object. Please update variable name.

integration-tests/downscope/downscope_test.go
```diff
@@ -136,15 +141,14 @@ func helper(tt downscopeTest) error {
...
 		return fmt.Errorf("failed to retrieve object from GCP project with error: %v", err)
 	}
-	dat, err := ioutil.ReadAll(obj.Body)
-	defer obj.Body.Close()
+	_, err = ioutil.ReadAll(obj.Body)
 	if err != nil {
 		return fmt.Errorf("ioutil.ReadAll: %v", err)
 	}
-
+	defer obj.Body.Close()
```
codyoss, line 152: This should called after the response nil check.

codyoss (changes requested):

integration-tests/downscope/downscope_test.go
```diff
@@ -141,14 +141,15 @@ func helper(t *testing.T, tt downscopeTest) error {
...
 	}
-	obj, err := storageService.Objects.Get(bucket, tt.objectName).Download()
+	resp, err := storageService.Objects.Get(bucket, tt.objectName).Download()
 	if err != nil {
 		return fmt.Errorf("failed to retrieve object from GCP project with error: %v", err)
 	}
-	_, err = ioutil.ReadAll(obj.Body)
+	_, err = ioutil.ReadAll(resp.Body)
 	if err != nil {
+		resp.Body.Close()
```
codyoss, line 150: Sorry if I was unclear, I meant to say this should be defered after line 147, before it has been read. That will cover all cases.

integration-tests/downscope/downscope_test.go
```diff
@@ -145,11 +145,10 @@ func downscopeQuery(t *testing.T, tt downscopeTest) error {
 	if err != nil {
 		return fmt.Errorf("failed to retrieve object from GCP project with error: %v", err)
 	}
+	defer resp.Body().Close()
```
codyoss, line 148: This does not compile. Should be `resp.Body.Close()`

Galadros: My bad, thanks.

codyoss (approved):

PR #1127, opened 2021-07-23
feat(internaloption): add AllowNonDefaultServiceAccount internaloption

mohanli-ml:
Add an internal option AllowNonDefaultServiceAccount for cloud services to bypass the check that credential must be retrieved from the metadata server and must use the default service account. In this way, non-default service account can be used for DirectPath.


codyoss:
@mohanli-ml Is there an AIP or design doc for this? I lack the context for this change.

broady:
@mohanli-ml Following up on Cody's comment... Is this something we can enable across all APIs? Why do we need an option for it?

mohanli-ml:
Sorry for the late response. This option is proposed to solve the conflict that DirectPath only support the default service account, while currently in GCP non-default service account is largely used. We:
1. collect some use cases in go/directpath-file-credentials;
2. have the proposal in go/directpath-gcp-authentication; 
3. have the rollout plan in go/directpath-file-credentials-rollout. 

This option should not be enabled before the service owners talk to the MDB group directpath-security-reviews and get their approval. See details in https://docs.google.com/document/d/1DgKCBcmz3J0OjbpO0BZBrwQIm3LrDBs8t9AS67NG0z4/edit#heading=h.qkiqy0p84mi8.

transport/grpc/dial.go
```diff
@@ -138,7 +138,7 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 		// * The endpoint is a host:port (or dns:///host:port).
 		// * Credentials are obtained via GCE metadata server, using the default
 		//   service account.
-		if o.EnableDirectPath && checkDirectPathEndPoint(endpoint) && isTokenSourceDirectPathCompatible(creds.TokenSource) && metadata.OnGCE() {
+		if o.EnableDirectPath && checkDirectPathEndPoint(endpoint) && isTokenSourceDirectPathCompatible(creds.TokenSource, o) && metadata.OnGCE() {
```
broady, line 141:
```
&& (o.AllowNonDefaultServiceAccount || isTokenSourceDirectPathCompatible(...)
``

broady:
LGTM, just a nit

transport/grpc/dial.go
mohanli-ml: Sorry it seems this comment will change the functionality. For example if `creds.TokenSource` is `nil`, then `isTokenSourceDirectPathCompatible(creds.TokenSource, o)` will be `false`, but `(o.AllowNonDefaultServiceAccount || isTokenSourceDirectPathCompatible(creds.TokenSource))` could be `true` if `o.AllowNonDefaultServiceAccount` is true. 

Based on my understanding, oauth token is per-RPC and attached to the RPC header, while service account is per-connection and attached in the ALTS context when DirectPath is doing ALTS handshake. So we should not change the functionality here.

broady (approved):

PR #1136, opened 2021-07-29
Change Active data type to pointer boolean

danielchristian-tokped:
Because golang didnt have null value, if a boolean variable's value is `false` it will be considered as `null` instead.
This could be quite problematic when someone wants to change or create a custom dimension with Inactive status because the `Active` field will instead be considered as an empty field an could the could cause an error.

To fix this, Active field will use pointer boolean instead so false value wouldn't be identified as `null`

### DIFF
--- a/analytics/v3/analytics-gen.go
+++ b/analytics/v3/analytics-gen.go
@@ -1326,7 +1326,7 @@ type CustomDimension struct {
 	AccountId string `json:"accountId,omitempty"`
 
 	// Active: Boolean indicating whether the custom dimension is active.
-	Active bool `json:"active,omitempty"`
+	Active *bool `json:"active,omitempty"`
 
 	// Created: Time the custom dimension was created.
 	Created string `json:"created,omitempty"`


codyoss:
Hey @danielchristian-tokped I think the solution to your problem is to use [ForceSendFields](https://pkg.go.dev/google.golang.org/api#hdr-ForceSendFields). Here is the field in the struct you are referencing:
https://github.com/googleapis/google-api-go-client/blob/6601d21aa6bc3891ebcc2b20c0b301c4b22a6d8f/analytics/v3/analytics-gen.go#L1373

danielchristian-tokped:
> Hey @danielchristian-tokped I think the solution to your problem is to use [ForceSendFields](https://pkg.go.dev/google.golang.org/api#hdr-ForceSendFields). Here is the field in the struct you are referencing:
> https://github.com/googleapis/google-api-go-client/blob/6601d21aa6bc3891ebcc2b20c0b301c4b22a6d8f/analytics/v3/analytics-gen.go#L1373

Ohh!! Thanks for telling me, my bad for not reading the documentation thoroughly that i missed that part on the documentation.
I'll close the Pull Request.

Issue #1055, opened 2021-06-05
false values for booleans are ignored unless included in ForceSendFields [admin API]


afumagalli:
#### Environment details

  - Programming language: Go
  - OS: macOS 11.4
  - Language runtime version: 1.13
  - Package version: 0.47.0

#### Steps to reproduce

  1. Make a Patch or Update to a user with `Suspension = true`
```
client.Users.Patch(userID, &admin.User{ Suspension: true }).Do()
```

  2. This will correctly suspend the user. Now do the opposite by making the same request with `Suspension = false`
```
client.Users.Patch(userID, &admin.User{ Suspension: false }).Do()
```

  3. The user will unexpectedly remain suspended. In order to get the un-suspension to apply you _must_ include `Suspension` in `ForceSendFields`
```
client.Users.Patch(userID, &admin.User{ Suspension: false, ForceSendFields: []string{ "Suspension" } }).Do()
```

It makes sense that `ForceSendFields` is required for nullable fields. But in this case the behavior is very surprising. In particular, the documentation in the package does not indicate that `ForceSendFields` would be required to send a `false` value

```
// ForceSendFields is a list of field names (e.g. "Alias") to
// unconditionally include in API requests. By default, fields with
// empty values are omitted from API requests. However, any non-pointer,
// non-interface field appearing in ForceSendFields will be sent to the
// server regardless of whether the field is empty or not. This may be
// used to include empty fields in Patch requests.
```

codyoss:
Hey, this is by design and fields like this are the reason while `ForceSendFields` exists. This is because all of the JSON structs include omitempty on fields. Hope that helps.

afumagalli:
Can the documentation at least be updated to illustrate this? The comment I pasted above is not clear that `ForceSendFields` is necessary to reason about `nil` vs `false`

codyoss:
@afumagalli That seems reasonable to me, this is also called out in the package docs: https://pkg.go.dev/google.golang.org/api#hdr-ForceSendFields

PR #1155, opened 2021-08-09
docs: clarify ForceSendFields default value behavior

codyoss:
Fixes: #1055


codyoss:
Client updates will happen with the next nightly generation to avoid conflicts with this PR.

PR #1162, opened 2021-08-12
feat(option): add internaloption to force use of certain credential

BrennaEpp:
A new internaloption added that forces the client to use a certain credential. This ensures that if a user directly passes in their own credentials, we can capture and use these credentials.

This is part of the effort to improve Signed URLs in the Go Storage Client.


option/internaloption/internaloption.go
```diff
@@ -106,3 +107,15 @@ type enableJwtWithScope bool
 func (w enableJwtWithScope) Apply(o *internal.DialSettings) {
 	o.EnableJwtWithScope = bool(w)
 }
+
+// WithCredentials returns a client option to specify authenticates API calls.
```
tritone, line 111: Seems like this sentence is a fragment-- should it be something like "a client option to specify credentials which will be used to authenticate API calls"?

tritone (approved):
Generally seems fine to me-- @codyoss @broady thoughts?

option/internaloption/internaloption.go
BrennaEpp: good catch!

codyoss:
LGTM, but I worked on this code so I will let @broady review.

broady (approved):

PR #1178, opened 2021-08-22
docs: corrects daysInCycle description & documents only accepted option

### LINKED ISSUE #1177: docs: incorrect documentaion for compute ResourcePolicyDailyCycle struct
The description for the `daysInCycle` property of compute's `ResourcePolicyDailyCycle` struct is incorrect. It reads:

 "Defines a schedule with units measured in **months**. The value determines how many **months** pass between the start of each cycle."
 
 Daily cycle resource policy schedules are measured in days and further are restricted to 1 day. Anything over and you would use a weekly policy and anything under you would use hourly. 
 
 Documentation should be changed to reflect Days instead of Months and should outline that the only accepted value for this property is `1`.

misterjacko:
fixes #1177 

There are also issues with the associated `compute-gen.go` files (including stating that they format is int64!) but there are instructions not to edit those files. 

I have changes to those files staged and can be added if needed.

### DIFF
--- a/compute/v0.alpha/compute-api.json
+++ b/compute/v0.alpha/compute-api.json
@@ -64424,7 +64424,7 @@
       "id": "ResourcePolicyDailyCycle",
       "properties": {
         "daysInCycle": {
-          "description": "Defines a schedule with units measured in months. The value determines how many months pass between the start of each cycle.",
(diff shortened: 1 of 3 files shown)


codyoss:
Hey @misterjacko thanks for the PR. Unfortunately the clients in this repo are auto generated so we can't make changes like this to the generated client directly. I will comment on the issue you raised about a good next step.

PR #1207, opened 2021-09-03
functions: Add ability to configure Google Container Filesystem (GCFS…

paulbsch:
…) on node pools.


codyoss:
Hey @paulbsch thanks for the PR, but this codebase is updated nightly automatically based on the public discovery doc for each API. If you would like to request a feature be added to an API I suggest reaching out on that teams support page. Hope that helps, thanks.

paulbsch:
Thanks for the response @codyoss .  I was not aware of the public discovery doc.  It is not readily obvious to me which team support page I should look for or how I would request this feature.  Where should I be looking?

codyoss:
Here is the discovery doc link for the file you were editing: https://container.googleapis.com/$discovery/rest?version=v1beta1

Here is a link to that teams support page: https://cloud.google.com/kubernetes-engine/docs/support

Hope that helps!

PR #1216, opened 2021-09-08
chore: Update CODEOWNERS for @googleapis/cloud-storage-dpe

danielbankhead:


tritone:
We don't need this here; the storage DPE team shouldn't have to be a codeowner.

PR #1224, opened 2021-09-14
feat(transport): Use `NewComputeEngineCredsWithOptions` to replace `NewComputeEngineCredentials`

mohanli-ml:
Use a new API `NewDirectPathCredentials` instead of `NewComputeEngineCredentials`. See context in grpc-go: https://github.com/grpc/grpc-go/pull/4767.


transport/grpc/dial.go
```diff
@@ -144,7 +144,7 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 			}
 			grpcOpts = []grpc.DialOption{
 				grpc.WithCredentialsBundle(
-					grpcgoogle.NewComputeEngineCredentials(),
+					grpcgoogle.NewComputeEngineCredentials(oauth.TokenSource{creds.TokenSource}),
```
codyoss, line 147: This code does not compile, this method does not take any arguments.

codyoss (changes requested):

transport/grpc/dial.go
mohanli-ml: Yes we will need to wait for grpc/grpc-go#4767. You may want to add a `do not merge` label for now.

PR #1260, opened 2021-10-19
feat(transport): add google-c2p dependence to DirectPath

mohanli-ml:
DirectPath is migrating to Traffic Director, and we want to setup E2E tests based on TD for Bigtable and Spanner. This PR:
1. Update grpc-go version to 1.41.0, which has RING_HASH LB policy support;
2. Add the google-c2p resolver dependence to google-api-go-client;
3. Add an environment variable GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS to enable using DirectPath over TD;
Corresponding PR in gax-java: https://github.com/googleapis/gax-java/pull/1521.

Also note that currently there is a bug in grpc-go, and fix is https://github.com/grpc/grpc-go/pull/4889. I have verified that with this fix Bigtable E2E test over DirectPath Traffic Director can pass. This PR has been merged to grpc-go and will release in grpc-go 1.42.0, which will be the minimal version for supporting `google-c2p` URI scheme.

### DIFF
--- a/abusiveexperiencereport/v1/abusiveexperiencereport-gen.go
+++ b/abusiveexperiencereport/v1/abusiveexperiencereport-gen.go
@@ -305,7 +305,7 @@ func (c *SitesGetCall) Header() http.Header {
 
 func (c *SitesGetCall) doRequest(alt string) (*http.Response, error) {
 	reqHeaders := make(http.Header)
-	reqHeaders.Set("x-goog-api-client", "gl-go/"+gensupport.GoVersion()+" gdcl/20211026")
+	reqHeaders.Set("x-goog-api-client", "gl-go/"+gensupport.GoVersion()+" gdcl/20211027")
 	for k, v := range c.header_ {
 		reqHeaders[k] = v
 	}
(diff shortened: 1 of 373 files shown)


tamird:
The change to transport/grpc/dial_enablexds.go in this PR causes a dependency on `google.golang.org/grpc/xds` and through it a huge ball of dependencies. Upgrading from v0.59.0 to v0.60.0 pulls in nearly **300k LOC** of dependencies. See https://gist.githubusercontent.com/tamird/56aff923e2b061546dc59bc7fec855cf/raw/4b84d173f9153d47fab555d35fa423b8e7edbf14/gistfile1.txt.

mohanli-ml:
Yes, adding xds support will explode dependencies and this size of google-api-go-client library, unfortunately. We have the same issue in other languages (for example, java https://github.com/googleapis/gax-java/pull/1521). I thought the dependencies are protected by the build tag, but looks like even if these codes are not compiled, they are still pulled?

tamird:
Yes, the Go module system can't know which build tags will be used, so it pulls all possible combinations.

tamird:
Filed https://github.com/googleapis/google-api-go-client/issues/1283.

PR #1275, opened 2021-10-28
fix: start reporting a meaningful version in headers

codyoss:


codyoss:
This is waiting for v3 of the action to fix a couple of bugs first.

codyoss:
Close in favor of #1426

PR #1297, opened 2021-11-16
feat(transport): pass call credentials to grpc-go for DirectPath

mohanli-ml:
Allow users to pass a call credential to grpc-go instead of always fetching the call credential from the metadata server, so that DirectPath can support file credentials. 

Problem: this PR relies on https://github.com/grpc/grpc-go/pull/4830, which is released in grpc-go 1.42. However, google-api-go-client is restricted to grpc-go 1.40, which should cause test to fail, and it seems build tag also can not help. I guess backport https://github.com/grpc/grpc-go/pull/4830 to grpc-go 1.40 is the only solution here? @menghanl


codyoss:
Yes @mohanli-ml , this should be backported if needed as we can't update to a newer version at this point.

menghanl:
Backporting a feature is not a common practice, but this might not be too bad since it's the `google` package.
It also means we would need to backport to 1.41. @dfawley WDYT?

And what's blocking updating gRPC to a newer version?

dfawley:
> And what's blocking updating gRPC to a newer version?

I assume this is Go version support.  @codyoss do you have any more visibility on when client libraries will be able to upgrade to require Go 1.13 or 1.14+?

I'm fine with backporting this to grpc-go 1.40 and 1.41.

codyoss:
@dfawley We are targeting Q1 22 right now, so hopefully not too far out now.

mohanli-ml:
@menghanl Hey Menghan, based on the discussion I think we will move forward with the backport to grpc-go 1.40 and 1.41? If so can you help with the backport? Thanks!

mohanli-ml:
@codyoss Hey Cody, we have just backported necessary support to grpc-go 1.41.1 and 1.40.1, https://github.com/grpc/grpc-go/releases. However, in google-api-go-client go.mod the grpc-go version is 1.40.0. Is it possible to increase it to 1.40.1?

codyoss:
@mohanli-ml I will get a PR right away!

mohanli-ml:
@codyoss Hey Cody, it seems the grpc version in go.mod is still v1.40.0?

mohanli-ml:
@codyoss Thanks for update the grpc version. Can you also help to re-trigger the test? I think it should pass now.

mohanli-ml:
@codyoss Hey Cody, tests passed and this PR should be ready for merge. Can you help to approve and merge it? Thanks!

codyoss (approved):

PR #1301, opened 2021-11-17
fix(dial): add nil checker

billysutomo:
When passing nil to option.ClientOption cause panic

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x1045db1e4]

goroutine 1 [running]:
google.golang.org/api/transport/http.newSettings({0x14000624180, 0x2, 0x2})
        /Users/billy/go/pkg/mod/google.golang.org/api@v0.60.0/transport/http/dial.go:105 +0x104
google.golang.org/api/transport/http.NewClient({0x1051f3e70, 0x140001a4000}, {0x14000624180, 0x2, 0x2})
        /Users/billy/go/pkg/mod/google.golang.org/api@v0.60.0/transport/http/dial.go:32 +0x38
google.golang.org/api/transport.NewHTTPClient(...)
        /Users/billy/go/pkg/mod/google.golang.org/api@v0.60.0/transport/dial.go:22
firebase.google.com/go/v4/messaging.NewClient({0x1051f3e70, 0x140001a4000}, 0x1400016c620)
        /Users/billy/go/pkg/mod/firebase.google.com/go/v4@v4.6.1/messaging/messaging.go:865 +0x54
firebase.google.com/go/v4.(*App).Messaging(0x140005f4120, {0x1051f3e70, 0x140001a4000})
        /Users/billy/go/pkg/mod/firebase.google.com/go/v4@v4.6.1/firebase.go:128 +0x88
main.main()
        /Users/billy/engineering/endorsaja/zeus/main.go:151 +0xf18
exit status 2
```

### DIFF
--- a/transport/http/dial.go
+++ b/transport/http/dial.go
@@ -102,7 +102,9 @@ func newTransport(ctx context.Context, base http.RoundTripper, settings *interna
 func newSettings(opts []option.ClientOption) (*internal.DialSettings, error) {
 	var o internal.DialSettings
 	for _, opt := range opts {
-		opt.Apply(&o)
+		if opt != nil {
+			opt.Apply(&o)
+		}
 	}
 	if err := o.Validate(); err != nil {
 		return nil, err


codyoss:
@billysutomo I am curious what the use case you ran into where this was uncovered. Generally if something requires an interface it is expected to be non-nil. I have no doubt that there are more places this type of check would need to be added if we wanted to add something like this.

billysutomo:
@codyoss actually I'm agree with you. There will be more nil checker for cases like that. Oke I'll decline my PR.
For the cases actually happen when i accidentally pass nil to initiate firebase app.
```
ctx := context.Background()
opt := option.WithCredentialsFile("file.json")
configFirebase := &firebase.Config{ProjectID: "endorsaja-28c6c"}
firebaseApp, err := firebase.NewApp(ctx, configFirebase, opt, nil) // here is the nil
_, err = firebaseApp.Messaging(ctx) // this trigger the panic
```

PR #1316, opened 2021-11-27
Update sheetID field . Must be non-negative

MomentoSystems:


codyoss (changes requested):
Hey, this is an autogeneated client so we don't accept contributions to it directly. The types here are omitted on purpose though because they are inferred by the compiler.

codyoss:
Hey, this is an autogeneated client so we don't accept contributions to it directly. The types here are omitted on purpose though because they are inferred by the compiler.

PR #1324, opened 2021-12-01
feat(internal/gensupport): add configurable retry

tritone:
Extend SendAndRetryRequest to allow retry options (backoff and
ErrorFunc) to be passed in. Add a surface to the generated code
for storage to allow the manual layer to choose whether the request
is retried.

Idempotency considerations will be handled at the manual layer.

I tested this out via the manual layer by verifying that the
ObjectsInsertCall only retries now when WithRetry() is applied.

I still need to update the generator to ensure that the changes in
the generated file will persist and probably add some tests as well.
Thought I'd try and get an initial review first.


google-api-go-generator/gen.go
```diff
@@ -2106,7 +2138,11 @@ func (meth *Method) generateCode() {
 		pn(`})`)
 	}
 	if meth.supportsMediaUpload() && meth.api.Name == "storage" {
-		pn("return gensupport.SendRequestWithRetry(c.ctx_, c.s.client, req)")
+		pn("if c.retry != nil {")
+		pn("	return gensupport.SendRequestWithRetry(c.ctx_, c.s.client, req, c.retry)")
+		pn("} else {")
```
noahdietz, line 2143: Nit: is the `else` necessary if the `if` always returns?

noahdietz (commented):
Seems fine to me, but I'll let @codyoss approve.

google-api-go-generator/gen.go
```diff
@@ -1986,6 +1992,32 @@ func (meth *Method) generateCode() {
...
+	if meth.supportsMediaUpload() && meth.api.Name == "storage" {
+		comment := "WithRetry causes the library to retry the initial request of the upload" +
+			"(for resumable uploads) or the entire upload (for multipart uploads) if" +
+			"a transient error occurs. This is contingent on ChunkSize being > 0 (so" +
+			"that the input data may be buffered). The backoff argument will be used to" +
+			"determine exponential backoff timing, and the errorFunc is used to determine" +
+			"which errors are considered retryable. By default, exponetial backoff will be" +
+			"applied using gax defaults, and the following errors are retried:" +
+			"\n\n" +
+			"- HTTP responses with codes 429, 502, 503, and 504." +
```
frankyn, line 2004: Could the error codes be pulled from gax  instead of string literals?

internal/gensupport/send.go
```diff
@@ -50,7 +52,7 @@ func send(ctx context.Context, client *http.Client, req *http.Request) (*http.Re
 // If ctx is non-nil, it calls all hooks, then sends the request with
 // req.WithContext, then calls any functions returned by the hooks in
 // reverse order.
-func SendRequestWithRetry(ctx context.Context, client *http.Client, req *http.Request) (*http.Response, error) {
+func SendRequestWithRetry(ctx context.Context, client *http.Client, req *http.Request, retry *RetryConfig) (*http.Response, error) {
```
frankyn, line 55: The only thing I think will be an issue is that the remote offset is not considered in the case Storage has progressed and the client is now unaligned.

frankyn (commented):
LGTM, just one main question on retries for resumable upload in general

google-api-go-generator/gen.go
tritone: Good point, fixed.

google-api-go-generator/gen.go
tritone: Seeing as this is just a one-off comment and these values are not likely to change, I don't think it's work adding extra complexity to the templating in order to accomplish this.

internal/gensupport/send.go
tritone: Yeah, that work will have to come in a separate PR. This PR doesn't change the mechanics of the resumption strategy.

codyoss (approved):
LGTM, thanks!

frankyn (approved):

PR #1337, opened 2021-12-06
chore: update gRPC version

codyoss:


codyoss:
cc @mohanli-ml

codyoss:
A feature was backported so we could update gRPC

noahdietz (approved):

Issue #1358, opened 2021-12-21
internal/gensupport: SendRequestWithRetry sometimes spuriously makes requests when its context was canceled


lhchavez:
#### Environment details

  - Programming language: Go (v1.7+)
  - OS: Any
  - Language runtime version: v1.7+
  - Package version: v0.61.0

#### Steps to reproduce

  1. Call `SendRequestWithRetry` with a canceled context
  2. The underlying network call _sometimes_ still happens. This is due to Go legally being allowed to visit either of the branches in the following snippet of code if both are satisfied: https://github.com/googleapis/google-api-go-client/blob/8fed2c8cba78a085e6e7ceda0fa5257170332ca1/internal/gensupport/send.go#L91-L100

Checking the context being done afterwards makes the issue go away.

#### Motivation

Some Cloud APIs expect callers to cancel the context if they want to cancel the operation, like
[`storage.(*ObjectHandle).NewWriter`](https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.NewWriter). The spuriousness of `SendRequestWithRetry` sometimes causes spooky things to happen, like uploading a 0-byte object to GCS, where no object was desired or expected!

PR #1359, opened 2021-12-21
fix(internal/gensupport): Make SendRequestWithRetry check for canceled contexts twice

lhchavez:
This change makes it deterministic to call SendRequestWithRetry with a
canceled context. This is going to be useful because some Cloud APIs
rely on the context being canceled to prevent some operations from
proceeding, like
[`storage.(*ObjectHandle).NewWriter`](https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.NewWriter).

Fixes: #1358

### DIFF
--- a/internal/gensupport/send.go
+++ b/internal/gensupport/send.go
@@ -99,6 +99,19 @@ func sendAndRetry(ctx context.Context, client *http.Client, req *http.Request, r
 		case <-time.After(pause):
 		}
 
+		select {
+		case <-ctx.Done():
+			// Check for context cancellation once more. If more than one case in a
+			// select is satisfied at the same time, Go will choose one arbitrarily.
+			// That can cause an operation to go through even if the context was
+			// canceled before.
+			if err == nil {
+				err = ctx.Err()
+			}
+			return resp, err
+		default:
+		}
+
 		resp, err = client.Do(req.WithContext(ctx))
 
 		var status int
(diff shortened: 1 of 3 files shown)


internal/gensupport/send_test.go
```diff
@@ -6,6 +6,7 @@ package gensupport
 
 import (
 	"context"
+	"errors"
```
codyoss, line 9: This module still supports 1.11 today. Could you please change this to xerrors.

internal/gensupport/send.go
```diff
@@ -99,6 +99,19 @@ func sendAndRetry(ctx context.Context, client *http.Client, req *http.Request, r
 		case <-time.After(pause):
 		}
 
+		select {
```
codyoss, line 102: Instead of a selecting I think we can just check if ctx.Err() is non-nil.

codyoss (changes requested):
Thanks for the PR and the detailed issue. Just a couple of small things.

internal/gensupport/send_test.go
lhchavez: changed!

internal/gensupport/send.go
lhchavez: yup, we totally can. TIL that `ctx.Err()` is threadsafe.

tritone (approved):
Great catch on this. Thanks for the fix!

tritone:
The presubmit is failing but if you run `go mod tidy` that should fix it. @lhchavez

lhchavez:
> The presubmit is failing but if you run `go mod tidy` that should fix it. @lhchavez 

tidied up!

codyoss (approved):
Thanks, look good to me!

Issue #1283, opened 2021-11-03
v0.60.0 pulls in 300kLOC of additional dependencies


tamird:
https://github.com/googleapis/google-api-go-client/commit/aa0f0be70f2d6003c6d25b0852ba754bc12cba34 added a dependency on google.golang.org/grpc/xds which pulls in a [lot of code](https://gist.githubusercontent.com/tamird/56aff923e2b061546dc59bc7fec855cf/raw/4b84d173f9153d47fab555d35fa423b8e7edbf14/gistfile1.txt).

It should be possible to upgrade past v0.60.0 without taking on all this bloat.

codyoss:
@mohanli-ml is there anything we can do here to lower the complexity of dependencies?

codyoss:
Can this resolver registration be moved to the site of an actual client?, or even better the program making use of our client? It seems like the bulk of the deps are related to pulling in a lot of the envoy project.

mohanli-ml:
@menghanl @dfawley Any suggestions here?

menghanl:
This import `_ "google.golang.org/grpc/xds/googledirectpath"` adds the extra dependencies.
(https://github.com/googleapis/google-api-go-client/commit/aa0f0be70f2d6003c6d25b0852ba754bc12cba34#diff-8a842d9ae06c25d12d6cf1e82d5057be6f81be7a2d75f82cc45aa28207f13023R16)

Direct path depends on xDS. To enable direct path, the library will need to pull in the xDS dependencies (including the xDS protos from the envoy project).

Is it OK for cloud libraries to _not_ have direct path capability by default, and ask users to enable direct path?
If the answer is yes, we can delete that import and ask the users to add it.
But note that it's a code change that they will need to make.

mohanli-ml:
Also if the decision is to move the xds dependency from the google-api-go-client repo to service repo (like bigtable-go, spanner-go), I want to know if we should also do this for other languages?

codyoss:
@mohanli-ml I think that it should at least be moved there. But I wonder if a better pattern would be to, through docs, tell users to add the registration import themselves. Then all users of spanner and bigtable don't have to pay the cost of these deps. I am not sure what the semantics look like in other languages, but if something similar could be done there I don't think that would be a bad idea.

dfawley:
[...]

PR #1361, opened 2021-12-21
feat(transport): remove google-c2p dependence to DirectPath

mohanli-ml:
Remove google-c2p dependency to the entire google-api-go-client. Instead, services using DirectPath should add the dependency by itself, like https://github.com/googleapis/google-cloud-go/pull/5090.

Issue: https://github.com/googleapis/google-api-go-client/issues/1283


transport/grpc/dial_enablexds.go
```diff
@@ -1,17 +0,0 @@
...
-// google.golang.org/api/option package to configure API clients.
-
-//go:build enablexds
-// +build enablexds
-
-package grpc
-
-import (
-	// Install google-c2p resolver, which is required for direct path.
-	_ "google.golang.org/grpc/xds/googledirectpath"
```
codyoss, line 16: Thanks, please run `go mod tidy`

codyoss (changes requested):

transport/grpc/dial_enablexds.go
mohanli-ml: Ah, sorry I forgot this. Just resubmitted. Thanks!

codyoss (approved):

PR #1364, opened 2021-12-22
fix(internal/gensupport): check ctx in chunk retry

tritone:
The select statement is non-deterministic, so currently, if the
pause is completed and ALSO the context has been canceled or
timeout elapsed, a request may still occur. This PR prevents that
circumstance from occurring.

Also removed something in a test that seems to be a workaround for
the race condition.

Inspired by #1359 & #1358.


internal/gensupport/resumable.go
```diff
@@ -177,6 +177,22 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
...
+			// That can cause an operation to go through even if the context was
+			// canceled before or the timeout was reached.
+			select {
+			case <-ctx.Done():
+				if err == nil {
+					err = ctx.Err()
+				}
+				return prepareReturn(resp, err)
+			case <-quitAfter:
+				return prepareReturn(resp, err)
```
codyoss, line 192: In this branch should we also check ctx.Err you thing? Or will err already be non-nil here?

codyoss (commented):
One question

internal/gensupport/resumable.go
tritone: It definitely should be non-nil; I don't think there is any way we could hit the per-chunk deadline (32s) without having at least one failed response first.

codyoss (approved):

PR #1399, opened 2022-01-15
fix: Adding import references in examples/customsearch

ZalgoNoise:
fix: Adding import references in examples/customsearch

Missing alias for the `customsearch` import statement will cause its references to raise errors. This type of package reference is also found in every other example, with similar import statements (`{package}/{version}`)


codyoss (approved):

ZalgoNoise:
Thank you!

PR #1403, opened 2022-01-18
chore: add new regen code

codyoss:
autosynth is going away later this week so we need to migrate the
regen. This code will eventually be called by a kokoro nightly job.


codyoss:
cc @SurferJeffAtGoogle for visability.

noahdietz (approved):

PR #1414, opened 2022-02-01
feat(gensupport): per-chunk deadline configs

### LINKED ISSUE #685: gensupport: resumable uploads shouldn't use a hardcoded retryDeadline
There is a hardcoded `retryDeadline` of 32s for each chunk of a resumable upload which seems rather short for reliable uploads: https://github.com/googleapis/google-api-go-client/blob/master/internal/gensupport/resumable.go#L27

When doing many resumable uploads to GCS we often encounter 503 responses for some objects lasting for longer than 32s, which makes the whole upload for that object fail.

It would be great if we could just use the deadline of the context for timing out retries or make the `retryDeadline` an option.

tritone:
feat(gensupport): per-chunk deadline configs

Allow users to configure the per-chunk deadline for retries
that's used during resumable uploads.

Needs to be exposed via the manual layer for storage.
    
Fixes #685

### DIFF
--- a/googleapi/googleapi.go
+++ b/googleapi/googleapi.go
@@ -15,6 +15,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"google.golang.org/api/internal/third_party/uritemplates"
 )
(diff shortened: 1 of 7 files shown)


codyoss (approved):
LGTM, just a question.

googleapi/googleapi.go
```diff
@@ -245,12 +246,30 @@ func ChunkSize(size int) MediaOption {
...
+
+// ChunkRetryDeadline returns a MediaOption which sets a per-chunk retry
+// deadline. If a single chunk has been attempting to upload for longer than
+// this time and the request fails, it will no longer be retried, and the error
+// will be returned to the caller.
+// This is only applicable for files which are large enough to require
+// a multi-chunk resumable upload.
+// The default value is 32s.
+// To set a deadline on the entire upload, use context timeout or cancellation.
+func ChunkRetryDeadline(deadline time.Duration) MediaOption {
```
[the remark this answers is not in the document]
tritone: Discussed offline, but-- yes, I think it does. Uploads can take a long time and you would want to know sooner if somehow you end up in a state where things can't continue. I think Jean and Emmanuel had considered removing this in the past but decided not to.

BrennaEpp (approved):
lgtm

googleapi/googleapi.go
```diff
@@ -245,12 +246,30 @@ func ChunkSize(size int) MediaOption {
...
+	o.ChunkRetryDeadline = time.Duration(cd)
+}
+
+// ChunkRetryDeadline returns a MediaOption which sets a per-chunk retry
+// deadline. If a single chunk has been attempting to upload for longer than
+// this time and the request fails, it will no longer be retried, and the error
+// will be returned to the caller.
+// This is only applicable for files which are large enough to require
+// a multi-chunk resumable upload.
+// The default value is 32s.
```
frankyn, line 261: Has 32s at per chunk been the default for a while now?

frankyn (changes requested):

googleapi/googleapi.go
codyoss: For 3ish years: https://github.com/googleapis/google-api-go-client/commit/4fac4de465568dff4e9b2a31b4d7a1da4ba2cefb

tritone: Yep, and there was a default of 16s prior to that apparently.

PR #1426, opened 2022-02-08
fix: start reporting a meaningful version in headers

codyoss:


internal/version.go
```diff
@@ -0,0 +1,7 @@
+// Copyright 2022 Google LLC. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package internal
+
+const Version = "0.67.0"
```
noahdietz, line 7: How will this be updated? Manually?

codyoss: Release please! https://github.com/googleapis/release-please/pull/1089

noahdietz: omg this would've been even better than that use of embed I had for showcase! Haha awesome.

PR #1428, opened 2022-02-08
feat: bump grpc and x/net

codyoss:
These are now unfrozen since min compat is now 1.15


codyoss:
cc @tritone I believe this unblocks you in some way once released... don't remember what though 😸

noahdietz (approved):

PR #1447, opened 2022-02-17
feat(transport): add an env variable to disable DirectPath

mohanli-ml:
Add an environment variable `GOOGLE_CLOUD_DISABLE_DIRECT_PATH` to disable DirectPath. We expect users to use the [EnableDirectPath option](https://github.com/googleapis/google-api-go-client/blob/main/option/internaloption/internaloption.go#L60) to control DirectPath for most cases, and this environment variable is for special cases like running both DP and GFE traffic on the same VM. As a result, only when this env variable is explicitly set to `true` will DirectPath be disabled.

Same PR for Java: https://github.com/googleapis/gax-java/pull/1412.

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -31,6 +31,9 @@ import (
 	_ "google.golang.org/grpc/balancer/grpclb"
 )
 
+// Check env to disable DirectPath traffic.
+const disableDirectPath = "GOOGLE_CLOUD_DISABLE_DIRECT_PATH"
+
 // Check env to decide if using google-c2p resolver for DirectPath traffic.
 const enableDirectPathXds = "GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS"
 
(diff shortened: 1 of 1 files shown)


codyoss:
@mohanli-ml is this a part of any [AIP](https://google.aip.dev/)? I think any envvar related to auth needs to be well documented for our users. Is there a plan for this?

mohanli-ml:
@codyoss Sorry we do not have a doc. 

The environment variable is not used for auth, its purpose is to provide service owners a more granular control of DirectPath. For now we use GCP project number ACL to control directPath, but we meet the situation where the same VM runs both CFE and DP probers. In this case ACL is not enough, and we need a more granular control. 

Only when this environment variable be set to "true" will DirectPath be disabled, otherwise the logic is the same as current.

codyoss (approved):
LGTM, let make sure we have a documentation plan for when this becomes more publicly available.

mohanli-ml:
Sure. Could you help to merge this PR? Thanks!

PR #1461, opened 2022-02-28
chore: remove build tags for go versions < 1.15

BrennaEpp:


codyoss (approved):
LGTM, thank you!

codyoss:
cc @quartzmo Here is an example where we used build tags to swap out impls in old versions of Go.

PR #1473, opened 2022-03-10
docs: removed stale comment

telpirion:
A customer noticed this comment in the documentation and became concerned about future support. Since this comment & feature have been in the library for > 6 years, can we get rid of the comment?


codyoss (approved):
Good catch. I agree, this is very stable at this point.

telpirion:
Thank you @codyoss !

PR #1476, opened 2022-03-14
feat(all): prefer using WithDefaultScope option internally

codyoss:
Currently these clients don't allow users to specify an aud because
both aud and scopes may not be provided at the same time. By moving
internal logic to use WithDefaultScopes we allow more flexability
with these clients. Updated local cached file.

Internal issue: b/223898679


codyoss:
Similar changes have been present in our GAPIC clients for a long time now: https://github.com/googleapis/google-cloud-go/pull/3374

noahdietz (approved):

PR #1506, opened 2022-04-09
feat(transport): remove grpc version guard

mohanli-ml:
Originally the gRPC version is because google-api-go-client can not depends on the latest version of gRPC, so we add the version guard and run tests that manually upgrade the gRPC version. Now that as google-api-go-client already depends on the latest gRPC, there is no need for this guard, so remove it.


codyoss (approved):
Thanks

PR #1525, opened 2022-04-28
feat: regen cloudcommerceprocurement v1

codyoss:
This is an out-of-band update as this API is no longer listed on the
public discovery index.


quartzmo (approved):
LGTM

PR #1554, opened 2022-05-22
use jsoniter

brachipa:


brachipa:
sorry, I mean to open it internally in my fork.

PR #1558, opened 2022-05-25
docs(idtoken): fix a typo on `NewClient` method

JohnTitor:
Should be "The token will have the audience...", I think.


codyoss (approved):
Thanks

PR #1570, opened 2022-06-04
feat(transport): Integrate with enterprise certificate proxy

andyrzhao:
Enterprise Certificate Proxy (https://github.com/googleapis/enterprise-certificate-proxy) is the new preferred solution for certificate based access on the client-side, since it integrates directly with the native OS keystores and does not expose the underlying private key. This new solution leverages the ECP client to replace the existing SecureConnect cert provider code-path for devices with the new ECP configuration file and helper binaries installed. (The ECP configuration file and helper binaries will eventually be distributed and managed through gCloud SDK). 

High-level change-list summary:
- Refactored default_cert.go to split out SecureConnectSource into it's own file.
- Added EnterpriseCertificateProxySource, a new client certificate source that uses the ECP client.
- Updated tests and comments and added more test coverage.
- Updated go.mod to include the new dependency.

Note that this change is safe and backwards compatible, since the ECP configuration file will not be present on devices without ECP binaries installed, in which case the DefaultSource will fall back to using SecureConnect. Furthermore, both the new ECP code-path and the old SecureConnect code-path are gated behind the env-var GOOGLE_API_USE_CLIENT_CERTIFICATE today.


transport/cert/enterprise_cert.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	return (&ecpSource{
+		key: key,
+	}).getClientCertificate, nil
+}
+
+func (s *ecpSource) getClientCertificate(info *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+	var cert tls.Certificate
+	cert.PrivateKey = s.key
```
shinfan, line 46: Do we need cache the key in this case? (like we did in the secure connect)

transport/cert/secureconnect_cert.go
```diff
@@ -0,0 +1,117 @@
...
+//
+// The configFilePath points to the location of the context aware metadata file.
+// If configFilePath is empty, use the default context aware metadata location.
+//
+// Return nil for Source and Error if config file is missing.
+func NewSecureConnectSource(configFilePath string) (Source, error) {
+	if configFilePath == "" {
+		user, err := user.Current()
+		if err != nil {
+			// Ignore.
```
shinfan, line 49: Can we comment why the error is ignored here? Same for L57

transport/cert/secureconnect_cert.go
```diff
@@ -0,0 +1,117 @@
...
+		return s.cachedCert, nil
+	}
+	// Expand OS environment variables in the cert provider command such as "$HOME".
+	for i := 0; i < len(s.metadata.Cmd); i++ {
+		s.metadata.Cmd[i] = os.ExpandEnv(s.metadata.Cmd[i])
+	}
+	command := s.metadata.Cmd
+	data, err := exec.Command(command[0], command[1:]...).Output()
+	if err != nil {
+		// TODO(cbro): read stderr for error message? Might contain sensitive info.
```
shinfan, line 96: Remove this TODO?

transport/cert/enterprise_cert.go
andyrzhao: The "key' in this case is actually a handle to the native keystore API, so there is nothing to cache. The secure connect code path actually works with a physical copy of the key, and the cache there prevents repeatedly executing the provider command to print out the key.

transport/cert/secureconnect_cert.go
andyrzhao: Done.

transport/cert/secureconnect_cert.go
andyrzhao: Done.

shinfan (approved):

shinfan:
@codyoss Could you please help review this change? Thanks!

transport/cert/testdata/context_aware_metadata.json
```diff
@@ -0,0 +1,3 @@
+{
+	"cert_provider_command":["cat", "testdata/testcert.pem"]
+}
```
codyoss, line 3: nit: a few of these files need newlines

codyoss (changes requested):
Just a few comments, or else looks good

transport/cert/default_cert.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC.
```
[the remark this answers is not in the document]
andyrzhao: Done.

transport/cert/enterprise_cert.go
```diff
@@ -0,0 +1,60 @@
...
+// a well-known gcloud location.
+//
+// Return nil for Source and Error if config file is missing, which
+// means Enterprise Certificate Proxy is not supported.
+func NewEnterpriseCertificateProxySource(configFilePath string) (Source, error) {
+	key, err := client.Cred(configFilePath)
+	if errors.Is(err, os.ErrNotExist) {
+		// Config file missing means Enterprise Certificate Proxy is not supported,
+		// so this is not a real error.
+		return nil, nil
```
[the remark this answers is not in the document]
andyrzhao: Added "errSourceUnavailable".

transport/cert/internal/test_signer.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2022 Google LLC.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+//
+// test_signer.go is a net/rpc server that listens on stdin/stdout, exposing
+// mock methods for testing enterprise certificate proxy flow.
+package main
```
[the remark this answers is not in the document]
andyrzhao: moved this to under transport/internal/ecp/test_signer.go

transport/cert/internal/test_signer.go
```diff
@@ -0,0 +1,104 @@
...
+	Opts   crypto.SignerOpts
+}
+
+// EnterpriseCertSigner exports RPC methods for signing.
+type EnterpriseCertSigner struct {
+	cert *tls.Certificate
+}
+
+// Transport wraps a pair of unidirectional streams as an io.ReadWriteCloser.
+type Transport struct {
```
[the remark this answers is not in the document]
andyrzhao: Done.

transport/cert/secureconnect_cert.go
```diff
@@ -0,0 +1,128 @@
...
+			return nil, nil
+		}
+		configFilePath = filepath.Join(user.HomeDir, metadataPath, metadataFile)
+	}
+
+	file, err := ioutil.ReadFile(configFilePath)
+	if errors.Is(err, os.ErrNotExist) {
+		// Config file missing means Secure Connect is not supported,
+		// so this is not a real error.
+		return nil, nil
```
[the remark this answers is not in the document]
andyrzhao: Done.

transport/cert/secureconnect_cert.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	var metadata secureConnectMetadata
+	if err := json.Unmarshal(file, &metadata); err != nil {
+		return nil, fmt.Errorf("cert: could not parse JSON in %q: %v", configFilePath, err)
```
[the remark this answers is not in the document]
andyrzhao: Done.

transport/cert/secureconnect_cert.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if err := validateMetadata(metadata); err != nil {
+		return nil, fmt.Errorf("cert: invalid config in %q: %v", configFilePath, err)
```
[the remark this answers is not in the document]
andyrzhao: Done.

transport/cert/testdata/context_aware_metadata.json
andyrzhao: Added EOF newlines to everything under testdata.

codyoss (commented):
It looks like a .DS_Store file was accidentally commited.

transport/cert/default_cert.go
```diff
@@ -49,93 +36,23 @@ var (
...
-// If that file does not exist, a nil source is returned.
+// If neither source is available (due to missing configurations), a nil Source and a nil Error are
+// returned to indicate that a default certificate source is unavailable.
 func DefaultSource() (Source, error) {
 	defaultCert.once.Do(func() {
-		defaultCert.source, defaultCert.err = newSecureConnectSource()
+		defaultCert.source, defaultCert.err = NewEnterpriseCertificateProxySource("")
+		if errors.Is(defaultCert.err, errSourceUnavailable) {
+			defaultCert.source, defaultCert.err = NewSecureConnectSource("")
+			if errors.Is(defaultCert.err, errSourceUnavailable) {
```
codyoss, line 52: I don' think this case3 is needed. In the case that err is non-nil and a source can't be detected we should just return the error.

transport/cert/enterprise_cert.go
```diff
@@ -0,0 +1,56 @@
...
+// certifcate related operations to an OS-specific "signer binary"
+// that communicates with the native keystore (ex. keychain on MacOS).
+//
+// The configFilePath points to a config file containing relevant parameters
+// such as the certificate issuer and the location of the signer binary.
+// If configFilePath is empty, the client will attempt to load the config from
+// a well-known gcloud location.
+func NewEnterpriseCertificateProxySource(configFilePath string) (Source, error) {
+	key, err := client.Cred(configFilePath)
+	if errors.Is(err, os.ErrNotExist) {
```
codyoss, line 38: nit: move this check into the err != nil check. If this is elsewhere consider doing the same.

transport/cert/secureconnect_cert_test.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"bytes"
+	"errors"
+	"testing"
+)
+
+func TestSecureConnectSource_ConfigMissing(t *testing.T) {
+	source, err := NewSecureConnectSource("missing.json")
+	if !errors.Is(err, errSourceUnavailable) {
+		t.Fatal("NewSecureConnectSource: expected errSourceUnavailable returned when config is missing.")
```
codyoss, line 15: nit: For all testing failures error messages should be in the form got then want.(opposite of many other langs) https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

codyoss (changes requested):
Just a coupe more small things.

transport/cert/default_cert.go
andyrzhao: So the reason why we return "nil, nil" when DefaultSource is unavailable is because the eventual caller of DefaultSource will take a different action when we don't have a certificate (i.e. it will not use mTLS path) - we don't want to error out in this scenario (in other words, nil Source is an acceptable return value). The other possibility to support this semantics is to export "errSourceUnavailable", so the caller can condition on that, but not sure if exporting custom errors is good practice, and it would also technically be non-backwards-compatible. WDYT?

transport/cert/secureconnect_cert_test.go
andyrzhao: Updated tests to be got then want. Thanks for your review and suggestions Cody!

transport/cert/default_cert.go
codyoss: You are right it would be a behavior change, but until this commit I don't think we documented that a nil error could be returned. It is definitely a grey area as far as breaking changes are concerned. But after thinking more I agree with you that although the current semantics are not ideal, I think it is best to keep them the same to avoid potential breakages.

codyoss (approved):

PR #1585, opened 2022-06-14
Google Takeout API

AlexDev404:


codyoss:
Hey, unfortunately we can't accept this PR. This project autogenerates client from a Googles public discovery doc index. If an API is not apart of this index our automation does not pick it up. If you would like to see this API added I suggest reaching out with an issue started from the support page of this API. Thanks for understanding.

PR #1640, opened 2022-08-02
chore: add flakybot

quartzmo:
Adapted from [google-cloud-go/kokoro/continuous.sh](https://github.com/googleapis/google-cloud-go/blob/main/internal/kokoro/continuous.sh#L168-L173).

closes: #1321

### DIFF
--- a/internal/kokoro/test.sh
+++ b/internal/kokoro/test.sh
@@ -56,3 +56,10 @@ if [[ $KOKORO_JOB_NAME == *"continuous"* ]]; then
 else
     go test -race -v -short ./... 2>&1 | tee $KOKORO_ARTIFACTS_DIR/$KOKORO_GERRIT_CHANGE_NUMBER.txt
 fi
+
+if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"continuous"* ]]; then
+  chmod +x $KOKORO_GFILE_DIR/linux_amd64/flakybot
+  $KOKORO_GFILE_DIR/linux_amd64/flakybot -logs_dir=$GOCLOUD_HOME \
+    -repo=googleapis/google-api-go-client \
+    -commit_hash=$KOKORO_GITHUB_COMMIT_URL_google_api_go_client
+fi


internal/kokoro/test.sh
```diff
@@ -56,3 +56,10 @@ if [[ $KOKORO_JOB_NAME == *"continuous"* ]]; then
 else
     go test -race -v -short ./... 2>&1 | tee $KOKORO_ARTIFACTS_DIR/$KOKORO_GERRIT_CHANGE_NUMBER.txt
 fi
+
+if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"continuous"* ]]; then
```
quartzmo, line 60: Do we only want to run flakybot on `continuous`, or should we always run it in this file, meaning for `go test -race -v -short ...` as well?

internal/kokoro/test.sh
```diff
@@ -56,3 +56,10 @@ if [[ $KOKORO_JOB_NAME == *"continuous"* ]]; then
...
+  chmod +x $KOKORO_GFILE_DIR/linux_amd64/flakybot
+  $KOKORO_GFILE_DIR/linux_amd64/flakybot -logs_dir=$GOCLOUD_HOME \
+    -repo=googleapis/google-api-go-client \
+    -commit_hash=$KOKORO_GITHUB_COMMIT_URL_google_api_go_client
```
quartzmo, line 64: Is `KOKORO_GITHUB_COMMIT_URL_google_api_go_client` valid in this context? I adapted this from [google-cloud-go](https://github.com/googleapis/google-cloud-go/blob/main/internal/kokoro/continuous.sh#L168-L173).

internal/kokoro/test.sh
codyoss: just continuous.

internal/kokoro/test.sh
codyoss: I believe it should be!

codyoss (approved):
Thanks!

PR #1650, opened 2022-08-09
chore: run gofmt -s -l -w . for Go 1.19

quartzmo:


codyoss (approved):

google-api-go-generator/gen_test.go
```diff
@@ -2,6 +2,9 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+//go:build go1.19
```
codyoss, line 5: Why is this needed?

quartzmo: The goldens for google-api-go-generator were failing for 1.19, so I updated them, and now the new goldens fail for the earliest version (and any other version < 1.19).

codyoss: Ah sure, since they include godoc. Interesting, would have never thought of that. Thanks!

PR #1663, opened 2022-08-16
feat(google-api-go-generator): Change field PaymentState to pointer

### LINKED ISSUE #727: androidpublisher: PaymentState should be a pointer type
Working with the api, I notice that PendingState is not using a pointer, but the value 0 is a valid information to business:
https://github.com/googleapis/google-api-go-client/blob/8001663557ac8d144131ea92d9829a0d8e668ab7/androidpublisher/v3/androidpublisher-gen.go#L1932

I believe will be best to change it to a pointer like the `PurchaseType` (example).

I notice that because when I call the `.MarshalJSON()` the field disappeared. 

If allowed, I can submit this PR.

eisandbar:
Added field PaymentState of schema SubscriptionPurchase of the androidpublisher:v3 api to the list of pointerFields. This was done the same way as it was previously done for v2.

Fixes #727


codyoss (approved):

codyoss:
Thanks!

PR #1665, opened 2022-08-16
feat(idtoken): Allow format options

### LINKED ISSUE #542: idtoken: computeEngine should support format and license options
id tokens set the format to `full` always which may include a lot of extra information in the token

eg. the gce instanceid, zone, etc, see https://github.com/googleapis/google-auth-library-nodejs/issues/792#issuecomment-575188336

and described [here](https://cloud.google.com/compute/docs/instances/verifying-instance-identity#token_format)

compute metadata based token uses by default `&format=standard` and does not include this extra info.  

FR to have an  option to allow setting the format and if it should include any license info (which is another parameter).    If possible, maybe make it default to standard but that may break existing users of the library.

Just note, there will be another GET parameter available soon which will allow including just the email value into the idtoken.  

the settings here hardcodes it to `full`

https://github.com/googleapis/google-api-go-client/blob/master/idtoken/compute.go#L41

eisandbar:
Added field CustomFormat to DialSettings in internal/settings.go
Added field format to computerIDTokenSource in idtoken/compute.go
Function computeTokenSource now sets field format to full, and if ds.CustomFormat != "" overwrites the field
Method Token now uses c.format instead of string literal "full"

Fixes #542


idtoken/compute.go
```diff
@@ -23,6 +23,10 @@ func computeTokenSource(audience string, ds *internal.DialSettings) (oauth2.Toke
 	}
 	ts := computeIDTokenSource{
 		audience: audience,
+		format:   "full",
```
codyoss, line 26: This changes the default behaviour which is not desired.

internal/settings.go
```diff
@@ -47,6 +47,7 @@ type DialSettings struct {
 	ImpersonationConfig           *impersonate.Config
 	EnableDirectPath              bool
 	AllowNonDefaultServiceAccount bool
+	CustomFormat                  string
```
codyoss, line 50: I don't think we want this plumbed to the DialSettings as this is a package things and not an over all client library thing.

Also, there is no mechanism to set this field. I think I have a pattern laying around somewhere for doing something like this but it involves exposing custom client options.

codyoss (changes requested):
Thanks for the contribution. I think I am going to hold off on this for now as we would need to figure out a way to limit the scope of these changes.

idtoken/compute.go
eisandbar: The default format was already full, so default behavior should be the same.

internal/settings.go
eisandbar: Right now DialSettings seems to already have a field called CustomClaims that only package idtoken uses

https://github.com/googleapis/google-api-go-client/blob/d6ee425a65668ee28ff97c6fb70f3497865d6572/internal/settings.go#L45

And the way it is set

https://github.com/googleapis/google-api-go-client/blob/62364a2da91e0116ee4520268eb660c2574629da/idtoken/idtoken.go#L146-L154

codyoss: You are right, but I don't think it should. This works but only because these two packages are in the same repo today, but in the future they may not be. DialSetting was used here out of convenience, but it muddies the waters of what DialSetting should be used for I think. 

There are patterns we could adopt to allow any package to declare its own options without needing to touch DialOptions at all. I think we CustomClaims could be re-written in such a manner -- as well as format.

eisandbar: Do you have a pattern in mind?

My first thought is have opts be of type interface{} and then type switch each opt
```go
func NewClient(ctx context.Context, audience string, opts ...interface{}) (*http.Client, error) {
	var ds internal.DialSettings
	for _, opt := range opts {
                switch x := opt.(type) {
                case ClientOption: 
                        x.Apply(&ds)
                }
	}
``` 

Are there any immediate drawbacks you see with this?

codyoss:
This is being implemented in our new auth library. Closing this PR as I don't think we will backport this. https://pkg.go.dev/cloud.google.com/go/auth/idtoken

PR #1666, opened 2022-08-16
fix(option): Having both a TokenSource and Scopes should fail validation

### LINKED ISSUE #1644: Specifying option.WithScopes is ignored when using option.WithTokenSource
If you use both

*  https://pkg.go.dev/google.golang.org/api/option#WithScopes
* https://pkg.go.dev/google.golang.org/api/option#WithTokenSource

at the same time as shown below, the scopes settings from the tokensource is used.

The following works for cloudidentity api because the "books" api scope is simply ignored

```golang
    configci, err := google.JWTConfigFromJSON(serviceAccountJSON, cloudidentity.CloudPlatformScope, cloudidentity.CloudIdentityGroupsReadonlyScope)

    tsci := configci.TokenSource(ctx)

    cisvc, err := cloudidentity.NewService(ctx, option.WithTokenSource(tsci), option.WithScopes("https://www.googleapis.com/auth/books"))
```

but if you invert it, you'll see a scopes error since the actual token from the config is used

```golang
    configci, err := google.JWTConfigFromJSON(serviceAccountJSON, "https://www.googleapis.com/auth/books")
    tsci := configci.TokenSource(ctx)

    cisvc, err := cloudidentity.NewService(ctx, option.WithTokenSource(tsci), option.WithScopes(cloudidentity.CloudPlatformScope, cloudidentity.CloudIdentityGroupsReadonlyScope))
```

consider throwing an error if both are set

---

```
$ go version
go version go1.17.1 linux/amd64
```

```golang
module main

go 1.17

require (
	golang.org/x/oauth2 v0.0.0-20220722155238-128564f6959c
	google.golang.org/api v0.91.0
)
```

eisandbar:
If both ds.TokenSource and ds.Scopes are set, then an error will be returned as Scopes will be ignored

Fixes #1644

### DIFF
--- a/option/option.go
+++ b/option/option.go
@@ -82,6 +82,9 @@ func (w withEndpoint) Apply(o *internal.DialSettings) {
 
 // WithScopes returns a ClientOption that overrides the default OAuth2 scopes
 // to be used for a service.
+//
+// If both WithScopes and WithTokenSource are used, scope settings from the
+// token source will be used instead.
 func WithScopes(scope ...string) ClientOption {
 	return withScopes(scope)
 }


codyoss (changes requested):
Although I think is a good change to make, it can't be made just yet. We will first need to audit our libraries to ensure we are not passing scopes via the WithScopes client option. I know where are in clients in this repo and may still have places where this is done in our sister repo as well. Adding this check would cause people to never be able to pass the TokenSource option right now in these instances.

I also due worry thinking more about this that there may be libraries built on libraries where these options are passed today. Once we have a better logging solution in place it may be better to log a warning in cases like this. I will need to think about this a little more.

eisandbar:
It was decided that returning an error is unnecessary and that simply clarifying the behavior in the docs will suffice.

PR #1670, opened 2022-08-17
docs(option): clarify behavior of WithScopes

### LINKED ISSUE #1644: Specifying option.WithScopes is ignored when using option.WithTokenSource

eisandbar:
Clarify that when both WithScopes and WithTokenSource are used, the
scope settings will be taken from the token source and the WithScopes
option will be ignored.

Fixes #1644

### DIFF
--- a/option/option.go
+++ b/option/option.go
@@ -82,6 +82,9 @@ func (w withEndpoint) Apply(o *internal.DialSettings) {
 
 // WithScopes returns a ClientOption that overrides the default OAuth2 scopes
 // to be used for a service.
+//
+// If both WithScopes and WithTokenSource are used, scope settings from the
+// token source will be used instead.
 func WithScopes(scope ...string) ClientOption {
 	return withScopes(scope)
 }


codyoss (approved):
Thank you for your contribution!

PR #1684, opened 2022-09-02
feat(internal/gensupport): wrap retry failures with context and prev error

matheuscscp:
I have a use case in which my context is cancelling a `sendAndRetry()` operation, and the error being returned is the error returned by the Google API in the previous attempt of `sendAndRetry()` instead of `ctx.Err()`, which is misleading because I was investigating why the Google Storage Go library was giving up on a 503 error (which is not expected according to the retry/idempotency docs: https://cloud.google.com/storage/docs/retry-strategy).

closes: #1685

### DIFF
--- a/internal/gensupport/send.go
+++ b/internal/gensupport/send.go
@@ -17,6 +17,27 @@ import (
 	"github.com/googleapis/gax-go/v2"
 )
 
+// Use this error type to return an error which allows introspection of both
+// the context error and the error from the service.
+type wrappedCallErr struct {
+	ctxErr     error
+	wrappedErr error
+}
+
+func (e wrappedCallErr) Error() string {
+	return fmt.Sprintf("retry failed with %v; last error: %v", e.ctxErr, e.wrappedErr)
+}
+
+func (e wrappedCallErr) Unwrap() error {
+	return e.wrappedErr
+}
+
+// Is allows errors.Is to match the error from the call as well as context
+// sentinel errors.
+func (e wrappedCallErr) Is(target error) bool {
+	return errors.Is(e.ctxErr, target) || errors.Is(e.wrappedErr, target)
+}
+
 // SendRequest sends a single HTTP request using the given client.
 // If ctx is non-nil, it calls all hooks, then sends the request with
 // req.WithContext, then calls any functions returned by the hooks in
@@ -43,7 +64,7 @@ func send(ctx context.Context, client *http.Client, req *http.Request) (*http.Re
 	if err != nil {
 		select {
 		case <-ctx.Done():
-			err = ctx.Err()
+			err = wrappedCallErr{ctx.Err(), err}
 		default:
 		}
 	}
@@ -98,11 +119,10 @@ func sendAndRetry(ctx context.Context, client *http.Client, req *http.Request, r
 		case <-ctx.Done():
 			// If we got an error, and the context has been canceled,
 			// the context's error is probably more useful.
-			retErr := ctx.Err()
-			if err != nil { // let's still return a previous error if any
-				retErr = fmt.Errorf("context error: %v. previous send error: %w", ctx.Err(), err)
+			if err != nil {
+				return resp, wrappedCallErr{ctx.Err(), err}
 			}
-			return resp, retErr
+			return resp, ctx.Err()
 		case <-time.After(pause):
 		}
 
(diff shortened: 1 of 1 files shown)


codyoss:
@matheuscscp could you please file an issue for discussion before I take a look at the PR. Thanks

matheuscscp:
> @matheuscscp could you please file an issue for discussion before I take a look at the PR. Thanks

@codyoss here it is: https://github.com/googleapis/google-api-go-client/issues/1685

sorry for not starting with that!

internal/gensupport/send.go
```diff
@@ -98,10 +98,11 @@ func sendAndRetry(ctx context.Context, client *http.Client, req *http.Request, r
 		case <-ctx.Done():
 			// If we got an error, and the context has been canceled,
 			// the context's error is probably more useful.
```
quartzmo, line 100: Would another possible solution to #1685 be simply to remove this entire block, and let the context raise the deadline exceeded error on its own?

matheuscscp: do you mean just `return nil, ctx.Err()`? otherwise, where would you suggest to put something equivalent to `case <-time.After(pause):` (the backoff)?

quartzmo: @tritone Can you comment on this?

tritone: Hey @quartzmo , thanks for flagging. I think the right answer is to return a wrapped error that includes both the context cancelation and the final error from the service, if available -- otherwise it can be pretty hard to debug either way. We implemented this in google-cloud-go: https://github.com/googleapis/google-cloud-go/blob/0da9ab0831540569dc04c0a23437b084b1564e15/internal/retry.go#L36

I can make a PR to implement this if folks agree.

tritone: (Also I see that this PR already basically takes that approach -- happy to work with you on completing it yourself if you want; the main change I would ask for is to implement a custom error type that implements `Is(error) bool` so that we can preserve the validity of calls to `errors.Is` in user code.

matheuscscp: yep I'd like to fix that as you suggested, I will do it tomorrow

tritone: sounds good, thank you!

matheuscscp: it's there, PTAL. I basically copied the `wrappedCallErr` struct from your other repo with the exception that I did `Is()` like this:

```
func (e wrappedCallErr) Is(target error) bool {
	return errors.Is(e.ctxErr, target) || errors.Is(e.wrappedErr, target)
}
```

basically leveraging `errors.Is()` instead of using plain `==` comparison. if that's too much pls let me know!

I also seized the opportunity to use this wrapper in all the other occurrences of some ctx+svc error

internal/gensupport/send.go
```diff
@@ -98,10 +119,10 @@ func sendAndRetry(ctx context.Context, client *http.Client, req *http.Request, r
 		case <-ctx.Done():
 			// If we got an error, and the context has been canceled,
```
tritone, line 120: This comment is no longer accurate, I would rephrase to something like:

```
// If we got an error and the context has been canceled, return an error acknowledging
// both the context cancelation and the service error.
```

internal/gensupport/send.go
```diff
@@ -43,7 +64,7 @@ func send(ctx context.Context, client *http.Client, req *http.Request) (*http.Re
 	if err != nil {
 		select {
 		case <-ctx.Done():
-			err = ctx.Err()
+			err = wrappedCallErr{ctx.Err(), err}
```
tritone, line 67: Let's not make this change in `send()` for now. Reasons:

1. The case that the call returned a non-context error but the context is expired/canceled is not a very common case without retries being in the picture. And the string from `wrappedCallErr` is specific to retries being at the root of the issue.
2. `send` is much more widely used than `sendAndRetry` including by the manual clients built on top of this library. I would want to audit these to make sure they can handle this error (meaning that they are upgraded to use `errors.Is` rather than checking for equality) if they have error handling logic at that layer.

tritone (commented):
Generally looks good, aside from change to `send` which I think we should hold off on.

internal/gensupport/send.go
matheuscscp: fixed!

internal/gensupport/send.go
matheuscscp: fixed!

tritone (approved):
LGTM. @codyoss could you also give this a look before we merge?

PR #1690, opened 2022-09-12
fix(gensupport): allow initial request for resumable uploads to retry w/ non-nil getBody

cojenco:
This allows retries when initiating a resumable upload session. 

Add support in `media.go` to return a `getBody` function so the request body can be recreated during retry attempts.

Tests pass locally in google-cloud-go/storage

Fixes https://github.com/googleapis/google-cloud-go/issues/6652


internal/gensupport/media.go
```diff
@@ -309,18 +308,30 @@ func (mi *MediaInfo) UploadRequest(reqHeaders http.Header, body io.Reader) (newB
...
-				_ = closer.Close()
-			}
-
-		}
 		reqHeaders.Set("Content-Type", ctype)
 		body = combined
 	}
 	if mi.buffer != nil && mi.mType != "" && !mi.singleChunk {
+		// This happens when initiating a resumable upload session.
+		// The initial request includes a body but no media. It can be
```
tritone, line 316: nit: I would say `// The initial request contains a JSON body rather than media.`

internal/gensupport/media_test.go
```diff
@@ -310,11 +310,11 @@ func TestUploadRequestGetBody(t *testing.T) {
 		},
 		{
 			desc: "chunk size < data size: MediaBuffer, >1 chunk, no getBody",
```
tritone, line 312: This should be "getBody" rather than "no getBody". Also, comment on L313 can be removed.

tritone (approved):
Two minor docs nits, otherwise LGTM

internal/gensupport/media.go
cojenco: Updated, thanks!

codyoss (approved):
LGTM

PR #1702, opened 2022-09-29
Fix typos in regex.

adob:
The original authors almost certainly intended to match the literal dash character '-' but ended up matching a range of other characters instead.


codyoss:
Unfortunetly we can't accept PRs for these files. They are pull from the public discovery doc listing. The source of truth would need to be updated, and those are not public. I would suggest filing issues on the associated services issue tracker in order to see a change made: https://cloud.google.com/support/docs/issue-trackers

Issue #1711, opened 2022-10-14
idtoken caching does not handle missing age in http response header


eytankidron:
In function calculateExpireTime (see https://github.com/googleapis/google-api-go-client/blob/977e871dbc6d880fa2392250a3d177598976b3d0/idtoken/cache.go#L98-L119) the code expects the "age" field to be present, otherwise the caching expiration is set to "now", which effectively means that the response is not cached.

According to https://httpwg.org/specs/rfc9111.html#age.calculations, a missing "age" field in the response header should be allowed and treated as if it was 0. So I think we should allow it to be missing.

codyoss:
This issue has been addressed and released.

PR #1715, opened 2022-10-17
fix(idtoken): Allow missing age in http response header

eytankidron:
Per https://github.com/googleapis/google-api-go-client/issues/1711 allow the "age" field to be missing from the http response header file and treat it as a zero if it is missing.


codyoss (approved):
Thanks!

PR #1729, opened 2022-10-28
fix: depend on newer version of compute

codyoss:
I believe this should alliviate the issue seen in
https://github.com/googleapis/google-cloud-go/issues/6956.


codyoss:
testing...

PR #1733, opened 2022-10-30
test(transport/grpc): fix arg for test failure logs

dragon3:
This pull request fixes arg for test failure logs in `transport/grpc/pool_test.go`


codyoss (approved):
Thanks!

Issue #1744, opened 2022-11-08
How to set a higher MaxIdleConnsPerHost for idtoken client


david-gang:
Thanks for stopping by to ask us a question! Please make sure to include:
- What you're trying to do

set a higher MaxIdleConnsPerHost for idtoken client  in order to get a better latency
- What code you've already tried

```
httpTransport := http.DefaultTransport.(*http.Transport).Clone()
	httpTransport.MaxIdleConnsPerHost = 100
	baseClient := http.Client{
		Transport: httpTransport,
	}
client, err := idtoken.NewClient(ctx, aud, option.WithHTTPClient(&baseClient))
```
- Any error messages you're getting

> transport/http: WithHTTPClient passed to NewTransport

The problem is similar to here:

https://github.com/googleapis/google-api-go-client/issues/1074

but the solution does not look applicable as i have no real access to the transport here.

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

quartzmo:
Hi @david-gang,

Regarding your sample:

```go
client, err := idtoken.NewClient(ctx, aud, option.WithHTTPClient(&baseClient))
```

It seems clear from the error that `option.WithHTTPClient` (or `idtoken.WithHTTPClient`) is not meant to be used with `idtoken.NewClient`. I think that in `idtoken` it is possibly only meant to be used with [idtoken.NewValidator](https://pkg.go.dev/google.golang.org/api/idtoken#NewValidator).

Is this a feature request to add `MaxIdleConnsPerHost` as an option?

david-gang:
Thanks @quartzmo !

PR #1754, opened 2022-11-18
fix(idtoken): increase MaxIdleConnsPerHost to 100 in NewClient

quartzmo:
closes: #1744

### DIFF
--- a/idtoken/idtoken.go
+++ b/idtoken/idtoken.go
@@ -54,7 +54,9 @@ func NewClient(ctx context.Context, audience string, opts ...ClientOption) (*htt
 	// Skip DialSettings validation so added TokenSource will not conflict with user
 	// provided credentials.
 	opts = append(opts, option.WithTokenSource(ts), internaloption.SkipDialSettingsValidation())
-	t, err := htransport.NewTransport(ctx, http.DefaultTransport, opts...)
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	httpTransport.MaxIdleConnsPerHost = 100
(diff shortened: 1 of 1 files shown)


idtoken/idtoken.go
```diff
@@ -54,7 +54,9 @@ func NewClient(ctx context.Context, audience string, opts ...ClientOption) (*htt
 	// Skip DialSettings validation so added TokenSource will not conflict with user
 	// provided credentials.
 	opts = append(opts, option.WithTokenSource(ts), internaloption.SkipDialSettingsValidation())
-	t, err := htransport.NewTransport(ctx, http.DefaultTransport, opts...)
+	httpTransport := http.DefaultTransport.(*http.Transport).Clone()
+	httpTransport.MaxIdleConnsPerHost = 100
```
codyoss, line 58: This seems like a large increase. The default when not set is 2 for comparison. I wonder if there is a different way we could enable users to do something like this.

quartzmo: This is the same default set in [transport/http/dial.go](https://github.com/googleapis/google-api-go-client/blob/main/transport/http/dial.go#L170). See some potentially relevant discussion [here](https://github.com/OJ/gobuster/issues/127).

quartzmo: @david-gang Do you have any opinion on hardcoding to `100` here, as opposed to flexible configuration or some other value?

david-gang: @quartzmo 100 is fine for me. In the product i work, we have 100 hardcoded for every http client.

codyoss (approved):

PR #1759, opened 2022-11-21
fix(transport/grpc): separate options for creds and certs

noahdietz:
For **gRPC**, this separates the logic for _credential_ dial options (associated with the `WithoutAuthentication` option) from the _connection-level encryption_ dial options (associated with the `WithInsecure` option OR using `DialInsecure`). This allows a _secure_ connection to be dialed _without_ credentials.

Previously, using `WithoutAuthentication` alone (i.e. without `WithInsecure` or `grpc.WithTransportCredentials(credentials.NewTLS(nil))`, and while using `Dial`) would result in attempting to dial a secure connection but without any credentials **or** TLS config, resulting in an error. For example, the following code returns an error:
```go
c, err := grpc.Dial(context.Background(),
  option.WithEndpoint("storage.googleapis.com:443"),
  option.WithoutAuthentication())
if err != nil {
  // grpc: no transport security set (use grpc.WithTransportCredentials(insecure.NewCredentials())
  // explicitly or set credentials)
}
```

This is not correct. It makes it very difficult for users to, for example, read a public GCS object without credentials, but using a secure connection.

This may seem like a weird use case, but the HTTP version of `dial` doesn't control connection-level encryption via the `WithoutAuthentication` option - connection certs loaded [here](https://github.com/googleapis/google-api-go-client/blob/main/transport/http/dial.go#L36-L44) while credentials are loaded [here](https://github.com/googleapis/google-api-go-client/blob/main/transport/http/dial.go#L74-L97).

Note: This also refactors the use of `grpc.WithInsecure`, which is no deprecated, to the preferred `grpcinsecure.NewCredentials()` option.

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	grpcgoogle "google.golang.org/grpc/credentials/google"
+	grpcinsecure "google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/credentials/oauth"
 
 	// Install grpclb, which is required for direct path.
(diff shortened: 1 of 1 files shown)


transport/grpc/dial.go
```diff
@@ -129,7 +129,15 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
...
-	} else if !o.NoAuth {
+	} else {
+		tlsConfig := &tls.Config{
+			GetClientCertificate: clientCertSource,
+		}
+		grpcOpts = []grpc.DialOption{
+			grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig)),
+		}
+	}
+	if !o.NoAuth {
```
noahdietz, line 140: One new behavior that would come of this refactor is that users could now accidentally send credentials over an insecure (unencrypted) connection. Should we force `o.NoAuth = true` when `insecure == true`? This would retain "backwards compatible" behavior for those using `WithoutAuthentication` but not `WithInsecure`. Not sure if these even worked in the first place because I get errors trying to use that configuration.

noahdietz: we _could_ just do `if !o.NoAuth && !insecure {` to retain the exact previous behavior.

codyoss: I think I like that idea. Can drop a todo in or file an issue to see if we should be more lenient in the future.

noahdietz: Ok, I've added this!

codyoss (approved):

PR #1775, opened 2022-12-07
chore: update golang.org/x/net to v0.4.0 to address CVE-2022-41717

thaJeztah:
golang.org/x/net contains a fix for CVE-2022-41717, which was addressed in stdlib in go1.19.4 and go1.18.9;

> net/http: limit canonical header cache by bytes, not entries
>
> An attacker can cause excessive memory growth in a Go server accepting
> HTTP/2 requests.
>
> HTTP/2 server connections contain a cache of HTTP header keys sent by
> the client. While the total number of entries in this cache is capped,
> an attacker sending very large keys can cause the server to allocate
> approximately 64 MiB per open connection.
>
> This issue is also fixed in golang.org/x/net/http2 v0.4.0,
> for users manually configuring HTTP/2.

full diff: https://github.com/golang/net/compare/f15817d10f9b...v0.4.0

other dependency updates:

- golang.org/x/sys v0.3.0: https://github.com/golang/sys/compare/3c1f35247d10...v0.3.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/googleapis/enterprise-certificate-proxy v0.2.0
 	github.com/googleapis/gax-go/v2 v2.7.0
 	go.opencensus.io v0.24.0
-	golang.org/x/net v0.0.0-20221014081412-f15817d10f9b
+	golang.org/x/net v0.4.0
 	golang.org/x/oauth2 v0.0.0-20221014153046-6fdb5e3db783
 	golang.org/x/sync v0.1.0
 	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2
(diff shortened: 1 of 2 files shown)


thaJeztah:
I'll do some follow-up PRs to update the other golang.org/x dependencies to tagged versions; the projects now tag SemVer releases, which probably would help with the renovate  bot also doing updates for them.

codyoss:
@thaJeztah Thanks for the report. I will need to do some digging here to see the implications of upgrading this and if our code is actually vulnerable to this report. The versions we use of the x repos today are a little older due to these libraries being compatible with newer versions of Go.

codyoss:
> Should the version advertised in go.mod...

This is unrelated to the version supported actually and only really enables using new lang features in this mod.(At least in todays Go ecosystem). Some of our packages might want to use newer lang features hidden behind features flags and things. There is actually an open issue about this confusion upstream Go somewhere....

thaJeztah:
Correct, that's what it _used_ to be, but with changes in how go modules are resolved based on that version, it's possible that module versions resolved in the go.mod are different based on what's in the file (TL;DR; it's becoming more and more ambiguous what the intent of the go version in go.mod is).

codyoss:
> but with changes in how go modules are resolved based on that version, it's possible that module versions resolved in the go.mod are different based on what's in the file

Yes, you are right about that. I believe that is why they have the compat flags if you need fully deterministic resolution with older versions. But yeah, a bit ambiguous indeed

thaJeztah:
Thanks for looking; no worries 👍

PR #1776, opened 2022-12-07
chore: remove uses of obsolete golang.org/x/xerrors

thaJeztah:
go1.13 and up now provide this functionality, and the xerrors.Is() function was deprecated.

Removing the dependency, as it was only used in a test.


codyoss (approved):
Thanks!

thaJeztah:
You're welcome! Was digging through our dependencies to figure out why this was still in our dependency tree (so ploughing my way through various repositories 😂)

PR #1782, opened 2022-12-14
feat: support set null map entries for non-simple map values

codyoss:
Previous impl assumed maps would be of type map[string]string. Although this is most of the map types this library uses some use more complex types. For these cases it is impossible to unset fields in patch requests until this fix. Reported internal

Ref: b/261221901


noahdietz (approved):

quartzmo (approved):

Sparrowej9:
Hi

PR #1789, opened 2022-12-16
fix(idtoken): configure validator constructor to use no authentication by default

adrianajg:


codyoss (approved):

codyoss:
@adrianajg When you fix a bug it is a good idea to put the bug in the body of the commit. Would you update the body to say

"Fixes: #XXXX". This also allows Github to cross-ref these two items

Issue #873, opened 2021-02-12
idtoken: support authorized_user credential type


bouk:
Right now there's no way to get an ID token if your credentials are user credentials e.g. if you're signed into gcloud locally. 

Right now I have to exec out into `gcloud auth print-identity-token` to get a token, it would be great if this library supported it natively.

bouk:
I managed to work around this, as follows:

```
import (
	"context"
	"fmt"

	"golang.org/x/oauth2"
	"golang.org/x/oauth2/google"
	"google.golang.org/api/idtoken"
)

func IDTokenTokenSource(ctx context.Context, audience string) (oauth2.TokenSource, error) {
	// First we try the idtoken package, which only works for service accounts
	ts, err := idtoken.NewTokenSource(ctx, audience)
	if err != nil {
		if err.Error() != `idtoken: credential must be service_account, found "authorized_user"` {
			return nil, err
		}
		// If that fails, we use our Application Default Credentials to fetch an id_token on the fly
		gts, err := google.DefaultTokenSource(ctx)
		if err != nil {
			return nil, err
		}
		ts = oauth2.ReuseTokenSource(nil, &idTokenSource{TokenSource: gts})
	}
	return ts, nil
}

// idTokenSource is an oauth2.TokenSource that wraps another
// It takes the id_token from TokenSource and passes that on as a bearer token
type idTokenSource struct {
	TokenSource oauth2.TokenSource
}

func (s *idTokenSource) Token() (*oauth2.Token, error) {
	token, err := s.TokenSource.Token()
	if err != nil {
		return nil, err
	}

	idToken, ok := token.Extra("id_token").(string)
	if !ok {
		return nil, fmt.Errorf("token did not contain an id_token")
	}

	return &oauth2.Token{
		AccessToken: idToken,
		TokenType:   "Bearer",
		Expiry:      token.Expiry,
	}, nil
}
```

This code will wrap the google oauth TokenSource to pass through its id_token

codyoss:
Hey @bouk you are correct that we don't currently support this, but this was done intentionally at the time of writing. The identity token spec we follow states this is optional: [AIP 4116](https://google.aip.dev/auth/4116). That being said I am curious if the other language libraries we support currently have this feature. @bshaffer do you know if user credentials are used in other languages?

bouk:
The python SDK seems to support it at least, since that's what gcloud uses under the hood (I assume?)

codyoss:
[...]

PR #1792, opened 2022-12-27
feat(idtoken): add support for impersonated_service_account creds type

adrianajg:
- Allow NewTokenSource function to accept authorization credentials of type `impersonated_service_account`

Updates: #873


idtoken/idtoken.go
```diff
@@ -67,6 +71,7 @@ func NewClient(ctx context.Context, audience string, opts ...ClientOption) (*htt
 // provided and configured with the supplied options. The parameter audience may
 // not be empty.
 func NewTokenSource(ctx context.Context, audience string, opts ...ClientOption) (oauth2.TokenSource, error) {
+	option.WithScopes()
```
codyoss, line 74: nit: this looks unintentional.

idtoken/idtoken.go
```diff
@@ -103,45 +108,72 @@ func newTokenSource(ctx context.Context, audience string, ds *internal.DialSetti
...
+		if err != nil {
+			log.Println(err)
+		}
+		return ts, nil
 	}
-	return oauth2.ReuseTokenSource(tok, ts), nil
 }
 
-func isServiceAccount(data []byte) error {
+// isOfAllowedType returns the credentials type as a string, and an error.
```
codyoss, line 160: nit: update docs to reflect new name

idtoken/idtoken.go
```diff
@@ -103,45 +108,72 @@ func newTokenSource(ctx context.Context, audience string, ds *internal.DialSetti
...
+// allowed types are "service_account" and "impersonated_service_account"
+func getAllowedType(data []byte) (string, error) {
```
codyoss, line 162: I wonder if instead this should return an enum(Go does not have such a thing but their are patterns). I think that would clean up this code a bit.

idtoken/idtoken.go
```diff
@@ -103,45 +108,72 @@ func newTokenSource(ctx context.Context, audience string, ds *internal.DialSetti
...
 	}
+	// if not service account return an error
 	if err := json.Unmarshal(data, &f); err != nil {
-		return err
+		return "", err
 	}
-	if f.Type != "service_account" {
-		return fmt.Errorf("idtoken: credential must be service_account, found %q", f.Type)
+	if f.Type != "service_account" && f.Type != "impersonated_service_account" {
+		return "", fmt.Errorf("idtoken: credential must be service_account or impersonated_service_account, found %q", f.Type)
```
codyoss, line 174: More types may be added over time, be less specific and maybe just say it is an unsupported type.

idtoken/idtoken.go
codyoss: If we did an enum I think this should not return an error, and then in the calling code handle the undefined enum state.

idtoken/idtoken.go
```diff
@@ -103,45 +108,72 @@ func newTokenSource(ctx context.Context, audience string, ds *internal.DialSetti
...
-	customClaims["target_audience"] = audience
-
-	cfg.PrivateClaims = customClaims
-	cfg.UseIDToken = true
-
-	ts := cfg.TokenSource(ctx)
-	tok, err := ts.Token()
-	if err != nil {
-		return nil, err
+	if allowedType == "service_account" {
```
codyoss, line 115: Feels like this should we a switch instead of an if/else

idtoken/idtoken_test.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2020 Google LLC.
```
codyoss, line 1: nit: update year

idtoken/idtoken_test.go
```diff
@@ -0,0 +1,80 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package idtoken
```
codyoss, line 5: I think this test should either be removed for now or else we need to turn it into a real integration test and setup creds. As is this is causing CI to fail.

codyoss (changes requested):

idtoken/idtoken.go
adrianajg: Removed.

codyoss:
@adrianajg It appears you have files that are not properly formatted

idtoken/integration_test.go
```diff
@@ -49,6 +50,58 @@ func TestNewTokenSource(t *testing.T) {
 	}
 }
 
+func TestNewTokenSource_WithImpersonatedServiceAccountCreds(t *testing.T) {
```
codyoss, line 53: Let remove this for now.

JohnsoNeal2013: Yea

JohnsoNeal2013:
Can someone please help me

adrianajg (commented):
Addressed changes, let me know if you see anything else. Thanks!

codyoss (approved):

PR #1799, opened 2023-01-04
chore: bump version of oauth lib to latest

liggitt:
cc @codyoss


codyoss:
All of the deps in this repo are behind for the moment. We plan on dropping support for some older versions of Go later this month, but until we do this as well as a couple other dependencies can not be upgraded due to deps relying on newer APIs. We currently still support 1.15.

codyoss:
Closing this as we will be bumping many deps together in open renovate PR.

Issue #1761, opened 2022-11-24
Memory Leak from time.After


lukemin89:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Is this a client library issue or a product issue? We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [Support page](https://cloud.google.com/support) to reach the most relevant engineers.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - Programming language: Golang
  - OS: MacOS
  - Language runtime version: 1.18
  - Package version: ?

#### Steps to reproduce

https://github.com/googleapis/google-api-go-client/blob/25e2e243ef14bc3ff3ecd47ac5172b87346a320f/internal/gensupport/resumable.go#L196

https://github.com/googleapis/google-api-go-client/blob/25e2e243ef14bc3ff3ecd47ac5172b87346a320f/internal/gensupport/resumable.go#L208

There are other places that use `time.After`.
If using with long `retryDeadline`, the memory usage keeps increasing.

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

quartzmo:
From: https://pkg.go.dev/time#After

> The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed.

codyoss:
cc @tritone

PR #1802, opened 2023-01-06
fix: use time.Timer over time.After

codyoss:
Can reduce memory usage for stacking timers that are not gc'ed in retry loops.

Fixes: #1761


quartzmo (approved):
LGTM

PR #1874, opened 2023-02-23
feat: add experimental s2a-go integration

xmenxk:
s2a integration with grpc and http transport code.
With this change, client library will try establishing mtls with the google api's mtls endpoint, using credentials held by s2a.
- also if dialing connection with s2a fails at runtime, it will fall back to dialing regular TLS to the regular endpoint.

The added code only changes the default behavior in  a way that's transparent to customers; if other mtls approach is specified, e.g. DCA, this code will not run.


quartzmo:
Hi @xmenxk,

Thank you for opening this pull request!

Can you please open a [feature request issue](https://github.com/googleapis/google-api-go-client/issues/new?assignees=&labels=type%3A+feature+request%2C+priority%3A+p3&template=feature_request.md) explaining the motivation for this feature (the benefits for users) and providing some background on the stability of [s2a-go](https://github.com/google/s2a-go) and how it will work in the context of this HTTP client?

quartzmo:
@xmenxk Is this PR related to #1886 in any way?

xmenxk:
> @xmenxk Is this PR related to #1886 in any way?

thanks @quartzmo for the response.

Both touch the transport layer code, as far as I can tell, they are for different use cases.

 #1886 looks like it's specifically for allowing configuring a http-client to talk to oauth2's mtls endpoint. 

The s2a integration this PR adds is a way for cloud workloads to talk to all google apis using MTLS, in a transparent way. It only runs when no other MTLS approach is specified, so there should't be any conflict.
I'll provide more information on s2a next week. thanks again!

transport/grpc/dial.go
```diff
@@ -8,26 +8,26 @@
...
 	"google.golang.org/api/option"
 	"google.golang.org/api/transport/internal/dca"
+	s2autil "google.golang.org/api/transport/internal/s2a"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	grpcgoogle "google.golang.org/grpc/credentials/google"
 	grpcinsecure "google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/credentials/oauth"
-
+	"log"
```
andyrzhao, line 27: run "gofmt -w" on the file to fix import ordering/formatting

transport/grpc/dial.go
```diff
@@ -135,6 +135,25 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 		transportCreds = credentials.NewTLS(&tls.Config{
 			GetClientCertificate: clientCertSource,
 		})
+		// try mtls with s2a when all the followings are true:
```
andyrzhao, line 138: so this is quite a bit of code to add and some duplicated code for the grpc & http implementations - can you encapsulate some of the logic into helper functions files? Also, as far as implementation goes, is there anyway to integrate s2a as a "clientCertSource" using the existing interface so it can be one of the options on the fallback path? (Both ECP enterprise-cert-proxy and secure-connect are on that code-path)

transport/internal/s2a/s2a.go
```diff
@@ -0,0 +1,12 @@
+package s2autil
+
+func MtlsEndpointIsEnabled(mtlsEndpoint string) bool {
```
andyrzhao, line 3: So how do you plan to maintain the source of truth for such an allowlist? Is the list specific to s2a or for all mTLS GCP apis in general? Note that the existing mTLS implementations in this library does not rely on such an allowlist - furthermore, the official mTLS endpoints used by this library are generated through what is known as the "discovery doc" tool chain

andyrzhao (changes requested):
Thank you for putting up this PR Kui! Left some general comments. I'm not familiar with the S2A architecture and would love to see some more context around this PR to make sure it is integrated smoothly.

transport/grpc/dial.go
xmenxk: Thanks Andy, I did some refactoring to move the code into separate files/functions, and it's cleaner now.

For the other question, I also thought about it initially, the issue is that S2A also implements other parts of the tls.Config struct, e.g. `VerifyPeerCertificate` method, so certificate source alone is not sufficient.
https://github.com/google/s2a-go/blob/36dafe55c3480837e3470a80970753cc63e0f3f5/internal/v2/tlsconfigstore/tlsconfigstore.go#L112

go.mod
```diff
@@ -18,10 +19,13 @@ require (
 	google.golang.org/protobuf v1.28.1
 )
 
+replace github.com/google/s2a-go => /usr/local/google/home/xmenxk/go/src/github.com/google/s2a-go
+
 require (
 	cloud.google.com/go/compute v1.18.0 // indirect
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
+	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519 // indirect
```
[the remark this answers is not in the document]
xmenxk: `god mod tidy` wants to keep it, I'll clean it in the end, thx

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.19
 require (
 	cloud.google.com/go/compute/metadata v0.2.3
 	github.com/google/go-cmp v0.5.9
+	github.com/google/s2a-go v0.0.0-20230207213108-7f75a7516db1
```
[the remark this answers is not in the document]
xmenxk: SG, will do:)

transport/internal/s2a/s2a.go
xmenxk: Thanks Andy, this was after chatting with Shin, we learned that even though the mtls endpoints are set in client libraries, sometimes it doesn't necessarily work. 
We are currently working with the team behind the `discovery` API to add an extra indicator (for each service) to mark whether an mtls endpoint is indeed MTLS enabled.  The plan is the default value will be `false` and only when we've confirmed with the service owner and verified it, it'll be set to true.

This function should mostly return that value from discovery doc. It can potentially take in a config (either from file or online) for overriding purpose. 

I'll share more details as we make progress on this. Plz let us know if you have any thoughts/suggestion on this.

transport/http/dial.go
```diff
@@ -41,7 +43,12 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*http.Client,
 	if settings.HTTPClient != nil {
 		return settings.HTTPClient, endpoint, nil
 	}
-	trans, err := newTransport(ctx, defaultBaseTransport(ctx, clientCertSource), settings)
+
+	dialTLSContext, targetEndpoint := s2autil.ConfigureHTTPTransport(clientCertSource, endpoint, settings)
```
andyrzhao, line 47: so if you update the helper to return the original endpoint when dialTLSContext is nil, you can assign to "endpoint" here directly.

On a related note, we should try to consolidate the endpoint calculation logic between internal/dca.go and s2util for the sake of consistency. If you take a look at https://github.com/googleapis/google-api-go-client/blob/main/internal/dca.go#L98 , the endpoint calculation is non trivial and factors in many things such "IS MTLS env var always/auto/never" and whether a certificateSource is present. s2a should work the same way and honor user requests to disable mTLS for instance (this impacts internal tests). I recommend you rename internal/dca.go to internal/cba.go (Certificate Based Access), add new helpers that to file such as internal.GetDialTLSContextAndEndpoint so you can reuse the endpoint calculation logic. If you consider s2a mTLS equivalent to device certs, then the logic should be parallel. WDYT?

andyrzhao (changes requested):

transport/http/dial.go
xmenxk: thanks I made some code consolidation  with dca.go per your suggestion.

internal/cba.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// Package dca contains utils for implementing Device Certificate
+// cba.go (certificate-based access) contains utils for implementing Device Certificate
 // Authentication according to https://google.aip.dev/auth/4114
```
andyrzhao, line 6: and Default Credentials for Google Cloud Virtual Environments according to https://google.aip.dev/auth/4115.

internal/cba.go
```diff
@@ -18,6 +18,10 @@
 //  2. If the user specifies an mTLS endpoint override but client certificate is not
 //     available, we will not fail-fast, but let backend throw error when connecting.
 //
+// If no DCA is specified, the code will try mTLS as default with credentials
+// held by the Secure Session Agent, which is part of Google's cloud infrastructure.
+// https://google.aip.dev/auth/4115
```
andyrzhao, line 23: you can omit link here if already mentioned at the top level comment

internal/cert/default_cert.go
```diff
@@ -44,7 +44,7 @@ var errSourceUnavailable = errors.New("certificate source is unavailable")
 //
 // If neither source is available (due to missing configurations), a nil Source and a nil Error are
 // returned to indicate that a default certificate source is unavailable.
-func DefaultSource() (Source, error) {
+var DefaultSource = func() (Source, error) {
```
andyrzhao, line 47: can you explain why the change here? If it is only to make it convenient for testing, then I'd rather we duplicate some code for testing instead of affecting prod execution path. Thanks!

internal/creds.go
```diff
@@ -92,7 +92,7 @@ func credentialsFromJSON(ctx context.Context, data []byte, ds *DialSettings) (*g
 
 	// Determine configurations for the OAuth2 transport, which is separate from the API transport.
 	// The OAuth2 transport and endpoint will be configured for mTLS if applicable.
-	clientCertSource, oauth2Endpoint, err := GetClientCertificateSourceAndEndpoint(oauth2DialSettings(ds))
+	clientCertSource, oauth2Endpoint, err := getClientCertificateSourceAndEndpoint(oauth2DialSettings(ds))
```
andyrzhao, line 95: Shouldn't this function be exported (i.e. capital G) for this to work?

andyrzhao (changes requested):
Looks good overall - I added some comments mostly around naming and code comments. Thanks! Also, please get one of the Yoshi team members to sign off on the change.

internal/creds.go
xmenxk: after #1886  all the callers are now within package `internal`, so it no longer needs to be exported

internal/cert/default_cert.go
xmenxk: this is just so that we can override it in unit tests for cases where we want no client cert to be available. 
On cloudtop for example, secureConnect is configured, so there will be cert source by default. 

let me know if you have other ideas. thanks!

xmenxk:
> Looks good overall - I added some comments mostly around naming and code comments. Thanks! Also, please get one of the Yoshi team members to sign off on the change.

Hi @codyoss, could you take a look at the PR? thanks :)

internal/cert/default_cert.go
andyrzhao: If that's the objective, you can simply override the os env var "GOOGLE_API_USE_CLIENT_CERTIFICATE" to "false" in your unit tests, and a nil cert source should be returned (make sure you restore the value to it's original value at the end of the test).

noahdietz:
> > Looks good overall - I added some comments mostly around naming and code comments. Thanks! Also, please get one of the Yoshi team members to sign off on the change.
> 
> Hi @codyoss, could you take a look at the PR? thanks :)

I've added @codyoss as a reviewer. I'm on rotation this week but I'd feel better with Cody reviewing this change anyways :)

andyrzhao (approved):
LGTM for CBA logic.

noahdietz:
Added cody back to reviewers

internal/cba.go
```diff
@@ -41,10 +52,12 @@ const (
 	mTLSModeAuto   = "auto"
 )
 
-// GetClientCertificateSourceAndEndpoint is a convenience function that invokes
+var useS2A = flag.Bool("use_s2a", false, "Experimental: if true, the code will try MTLS with S2A as the default for transport security.")
```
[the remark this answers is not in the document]
codyoss: I agree that I think a envvar would be preferable here.

internal/cba.go
```diff
@@ -56,6 +70,40 @@ func GetClientCertificateSourceAndEndpoint(settings *DialSettings) (cert.Source,
 	return clientCertSource, endpoint, nil
 }
 
+// getTransportConfig returns four values:
+//
+//	the client certificate source,
+//	the corresponding endpoint to use based on client certificate source,
+//	the S2A address if it can be used, otherwise returns empty string,
+//	the endpoint to use with S2A.
+func getTransportConfig(settings *DialSettings) (cert.Source, string, string, string, error) {
```
codyoss, line 79: nit: consider a struct return type here to improve readability and usage.

codyoss (commented):
LGTM overall, couple of small things

internal/cba.go
xmenxk: done thanks!

internal/cba.go
```diff
@@ -0,0 +1,282 @@
...
+	"google.golang.org/grpc/credentials"
+)
+
+const (
+	mTLSModeAlways = "always"
+	mTLSModeNever  = "never"
+	mTLSModeAuto   = "auto"
+
+	// Experimental: if true, the code will try MTLS with S2A as the default for transport security. Default value is false.
+	googleAPIUseS2AEnv = "GOOGLE_API_USE_S2A"
```
codyoss, line 55: nit: can you add the work `EXPERIMENTAL` some where in this envvar. Should we stabilize it in the future we can change it and get it properly documented

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.19
 require (
 	cloud.google.com/go/compute/metadata v0.2.3
 	github.com/google/go-cmp v0.5.9
+	github.com/google/s2a-go v0.1.0
```
liggitt, line 8: this is unfortunate... it adds more transitive dependencies on appengine and means more consumers of this module now link in appengine code - xref https://github.com/kubernetes/kubernetes/pull/120341#discussion_r1312959575

liggitt: can use of this be isolated in a different package, and/or build-tagged so all consumers are not exposed to an experimental / unstable transitive dependency?

quartzmo: For example, https://github.com/google/s2a-go/blob/3d294feadf626dd327a2cbba9e91d913aefbdea3/internal/handshaker/service/service.go#L31

xmenxk: We should be able to drop the appengine depdency from s2a-go repo.

xmenxk: synced with @liggitt 
removing appengine dependency from s2a in https://github.com/google/s2a-go/pull/119

Issue #1879, opened 2023-02-27
Support idtoken for `external_account` type (Workload Identity Federation)


senthilkumarkj:
`external_account` type is used for keyless workload identity described here - https://google.aip.dev/auth/4117

go client supports generating access token for this type - https://pkg.go.dev/golang.org/x/oauth2/google#hdr-Workload_Identity_Federation

ID token generation does not work yet - please refer https://github.com/googleapis/google-api-go-client/issues/1393#issuecomment-1154046913

senthilkumarkj:
This is fixed in PR https://github.com/googleapis/google-api-go-client/pull/1897 and available in v0.113.0 (https://github.com/googleapis/google-api-go-client/pull/1901)

PR #1897, opened 2023-03-08
feat(idtoken): add support for external_account

senthilkumarkj:
* Also fix a bug for impersonated_service_account * When creds are passed with WithCredentialsFile(), it doesn't work. * Pass the option when creating the token source.
* Fixes issue #1879


senthilkumarkj:
JFYI, below is the test file I used to test with both GOOGLE_APPLICATION_CREDENTIALS and --service_account_key_path flag. https://gist.githubusercontent.com/senthilkumarkj/7ba55c295c9bd58e02c47d8f2c3b508d/raw/05c05d70cab4d7e159bb6c0fca50cbb7b410285f/idtoken.go

codyoss (approved):

idtoken/idtoken.go
```diff
@@ -34,6 +34,7 @@ const (
 	unknownCredType credentialsType = iota
 	serviceAccount
 	impersonatedServiceAccount
+	external_account
```
codyoss, line 37: nit: can you please change this to `externalAccount`

senthilkumarkj: good catch, I totally overlooked it! Fixed now.

PR #1916, opened 2023-03-20
chore(gensupport): add idempotency header

tritone:
This header will allow GCS object uploads to be idempotent by default. For other products, it will have no effect.


noahdietz (dismissed):

internal/gensupport/resumable.go
```diff
@@ -85,6 +85,9 @@ func (rx *ResumableUpload) doUploadRequest(ctx context.Context, data io.Reader,
 	invocationHeader := fmt.Sprintf("gccl-invocation-id/%s gccl-attempt-count/%d", rx.invocationID, rx.attempts)
```
noahdietz, line 85: Out of curiosity, will this extra `gccl-invocation-id` key-value pair still be necessary in `X-Goog-Api-Client` if there is a purpose-built header now? If not, please file an issue and add it as a todo to remove this later.

noahdietz (approved):

PR #2008, opened 2023-05-31
chore: remove refs to deprecated io/ioutil

testwill:


codyoss (approved):
Thanks!

google-api-go-generator/gen.go
```diff
@@ -329,15 +328,15 @@ func isNewerRevision(old []byte, new []byte) error {
 
 func writeFile(file string, contents []byte) error {
 	// Don't write it if the contents are identical.
-	existing, err := ioutil.ReadFile(file)
+	existing, err := os.ReadFile(file)
```
codyoss, line 331: Should be `io` here

codyoss (changes requested):

PR #2103, opened 2023-08-07
feat: additional checks before using S2A

xmenxk:
add a couple more checks before using S2A for mtls
1. make sure `DefaultMTLSEndpoint` is not empty
2. skip s2a if custom http client is provided, because there's no guarantee the client passed in is mtls-capable

### DIFF
--- a/internal/cba.go
+++ b/internal/cba.go
@@ -91,16 +91,10 @@ func getTransportConfig(settings *DialSettings) (*transportConfig, error) {
 		s2aMTLSEndpoint:  "",
 	}
 
-	// Check the env to determine whether to use S2A.
-	if !isGoogleS2AEnabled() {
+	if !shouldUseS2A(clientCertSource, settings) {
 		return &defaultTransportConfig, nil
 	}
 
-	// If client cert is found, use that over S2A.
-	// If MTLS is not enabled for the endpoint, skip S2A.
-	if clientCertSource != nil || !mtlsEndpointEnabledForS2A() {
-		return &defaultTransportConfig, nil
-	}
 	s2aMTLSEndpoint := settings.DefaultMTLSEndpoint
 	// If there is endpoint override, honor it.
 	if settings.Endpoint != "" {
@@ -116,10 +110,6 @@ func getTransportConfig(settings *DialSettings) (*transportConfig, error) {
 		s2aAddress:       s2aAddress,
 		s2aMTLSEndpoint:  s2aMTLSEndpoint,
 	}, nil
-}
-
-func isGoogleS2AEnabled() bool {
-	return strings.ToLower(os.Getenv(googleAPIUseS2AEnv)) == "true"
 }
 
 // getClientCertificateSource returns a default client certificate source, if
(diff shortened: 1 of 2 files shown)


go.work.sum
```diff
@@ -1,13 +1,286 @@
 cloud.google.com/go v0.110.0 h1:Zc8gqp3+a9/Eyph2KDmcGaPtbKRIoqq4YTlL4NMD0Ys=
 cloud.google.com/go v0.110.2 h1:sdFPBr6xG9/wkBbfhmUz/JmZC7X6LavQgcrVINrKiVA=
 cloud.google.com/go v0.110.4 h1:1JYyxKMN9hd5dR2MYTPWkGUgcoxVVhg0LKNKEo0qvmk=
+cloud.google.com/go v0.110.4/go.mod h1:+EYjdK8e5RME/VY/qLCAtuyALQ9q67dvuum8i+H5xsI=
```
clundin25, line 4: I am not familiar with this code base, but should this file be updated separately?

Are all of these dependencies being updated?

internal/cba.go
```diff
@@ -97,8 +97,11 @@ func getTransportConfig(settings *DialSettings) (*transportConfig, error) {
 	}
 
 	// If client cert is found, use that over S2A.
+	// If DefaultMTLSEndpoint is not set, skip S2A.
 	// If MTLS is not enabled for the endpoint, skip S2A.
-	if clientCertSource != nil || !mtlsEndpointEnabledForS2A() {
+	// If custom HTTP client is provided, skip S2A.
+	if clientCertSource != nil || !mtlsEndpointEnabledForS2A() ||
+		settings.DefaultMTLSEndpoint == "" || settings.HTTPClient != nil {
```
clundin25, line 104: IMO this conditional is complicated enough that it should be wrapped by a function, maybe something like this:

```golang
func ShouldUseS2A($WHATEVER_PARAMS_YOU_NEED) bool
```

What do you think?

go.work.sum
noahdietz: Can you just revert this, update your branch and double check that everything is in order? I might need to just manually update this on `main` so that you don't need to change this.

xmenxk: Thanks @noahdietz I reverted changes to this file. PTAL

internal/cba.go
```diff
@@ -275,8 +265,25 @@ func GetHTTPTransportConfigAndEndpoint(settings *DialSettings) (cert.Source, fun
...
 
+func shouldUseS2A(clientCertSource cert.Source, settings *DialSettings) bool {
+	// If client cert is found, use that over S2A.
+	// If EXPERIMENTAL_GOOGLE_API_USE_S2A is not set to true, skip S2A.
+	// If DefaultMTLSEndpoint is not set, skip S2A.
+	// If MTLS is not enabled for this endpoint, skip S2A.
+	// If custom HTTP client is provided, skip S2A.
+	if clientCertSource != nil || !isGoogleS2AEnabled() || settings.DefaultMTLSEndpoint == "" ||
+		!mtlsEndpointEnabledForS2A() || settings.HTTPClient != nil {
+		return false
```
clundin25, line 276: nit: I would consider unrolling this to multiple conditionals to improve the readability. Given there are 5 boolean expressions, and two of them with a negation, it is hard to follow the logic.

clundin25:
@xmenxk can you fix the conventional commit error?

internal/cba.go
xmenxk: done thanks for the comments!

codyoss (approved):

rwuorinen81:
CAN ANYONE HELP ME ACCESS MY WIFE THAT PASSED AWAYS YAHOO EMAIL ??)PLEASE

PR #2126, opened 2023-08-30
docs: point more users forward

codyoss:
Some users still land on our older libraries first, lets put a similar disclaimer that is in our README into all our generated client godoc.

Also cleaned up some formatting of these package docs.

Internal bug: 288938991


noahdietz (approved):

rwuorinen81:
My wife's phone got stolen and we can't access her back up security  when
she gets sent a code...on gmail..what do we do ??

PR #2136, opened 2023-09-07
feat(idtoken): add ParsePayload returning unvalidated token payload

### LINKED ISSUE #2137: Add a function idtoken.GetPayload()
idtoken.GetPayload() should be similar to idtoken.Validate() except it skips the validation part and only returns the payload if it can parse it.

The reason for this function is to better be able to debug cases where idtoken.Validate() fails and we still want to know what was in the payload. Note that idtoken.Validate() returns the payload on success but not on failure.

I created https://github.com/googleapis/google-api-go-client/pull/2136 to implement this issue. This PR also includes a slightly more informative error message when the token is expired.

eytankidron:
`GetPayload` attempts to parse the token and return the parsed payload from therein. It does not attempt to `Validate` the token semantically or otherwise. This is especially useful for the case where `Validate` _fails_ and the user wants to inspect the payload directly (`Validate` doesn't return the payload on failure currently).

Also, improve the error message for token expiration by including the expiration time and the evaluated time.

Fixes #2137


codyoss:
Please open an issue for discussion before making changes detailing the how/why for the feature request.

eytankidron:
Created https://github.com/googleapis/google-api-go-client/issues/2137

idtoken/validate.go
```diff
@@ -122,6 +122,34 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
 	return defaultValidator.validate(ctx, idToken, audience)
 }
 
+// GetPayload just gets the payload part of the token.
```
noahdietz, line 125:
```suggestion
// GetPayload parses the given token and returns its payload.
```

idtoken/validate.go
```diff
@@ -122,6 +122,34 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
...
+// this function returns.
+func (v *Validator) GetPayload(ctx context.Context, idToken string) (*Payload, error) {
+	jwt, err := parseJWT(idToken)
+	if err != nil {
+		return nil, err
+	}
+	return jwt.parsedPayload()
+}
+
+// GetPayload just gets the payload part of the token.
```
noahdietz, line 141:
```suggestion
// GetPayload parses the given token and returns its payload.
```

idtoken/validate.go
```diff
@@ -122,6 +122,34 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
 	return defaultValidator.validate(ctx, idToken, audience)
 }
 
+// GetPayload just gets the payload part of the token.
+//
+// WARNING: THIS FUNCTION DOES NOT VALIDATE THE TOKEN.
```
noahdietz, line 127:
```suggestion
// Warning: This function does not validate the token prior to parsing it.
```

idtoken/validate.go
```diff
@@ -122,6 +122,34 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
...
+	jwt, err := parseJWT(idToken)
+	if err != nil {
+		return nil, err
+	}
+	return jwt.parsedPayload()
+}
+
+// GetPayload just gets the payload part of the token.
+//
+// WARNING: THIS FUNCTION DOES NOT VALIDATE THE TOKEN.
```
noahdietz, line 143:
```suggestion
// Warning: This function does not validate the token prior to parsing it.
```

idtoken/validate.go
```diff
@@ -122,6 +122,34 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
...
 }
 
+// GetPayload just gets the payload part of the token.
+//
+// WARNING: THIS FUNCTION DOES NOT VALIDATE THE TOKEN.
+//
+// In fact, it explicitly skips the validation part. It should only be used to inspect the payload
+// content of a token, perhaps for debugging purposes, as a means to try to figure out why the
+// validation failed. Note that if Validate() succeeds, it already returns the exact payload that
+// this function returns.
```
noahdietz, line 132:
```suggestion
// GetPayload is primarily meant to be used to inspect a token's payload. This is
// useful when validation fails and the payload needs to be inspected.
//
// Note: A successful Validate() invocation with the same token will return an
// identical payload.
```

idtoken/validate.go
```diff
@@ -122,6 +122,34 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
...
+}
+
+// GetPayload just gets the payload part of the token.
+//
+// WARNING: THIS FUNCTION DOES NOT VALIDATE THE TOKEN.
+//
+// In fact, it explicitly skips the validation part. It should only be used to inspect the payload
+// content of a token, perhaps for debugging purposes, as a means to try to figure out why the
+// validation failed. Note that if Validate() succeeds, it already returns the exact payload that
+// this function returns.
```
noahdietz, line 148:
```suggestion
// GetPayload is primarily meant to be used to inspect a token's payload. This is
// useful when validation fails and the payload needs to be inspected.
//
// Note: A successful Validate() invocation with the same token will return an
// identical payload.
```

noahdietz (changes requested):
Please add some tests as well, thanks.

idtoken/validate.go
eytankidron: Done.

idtoken/validate.go
eytankidron: Done.

idtoken/validate.go
eytankidron: Done.

idtoken/validate.go
eytankidron: Done.

idtoken/validate.go
eytankidron: Done.

idtoken/validate.go
eytankidron: Done.

eytankidron:
Thanks Noah for the comment improvements. I will also add unit tests shortly.

idtoken/validate.go
```diff
@@ -122,6 +122,36 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
...
+// GetPayload parses the given token and returns its payload.
+//
+// Warning: This function does not validate the token prior to parsing it.
+//
+// GetPayload is primarily meant to be used to inspect a token's payload. This is
+// useful when validation fails and the payload needs to be inspected.
+//
+// Note: A successful Validate() invocation with the same token will return an
+// identical payload.
+func (v *Validator) GetPayload(ctx context.Context, idToken string) (*Payload, error) {
```
codyoss, line 134: nit: because no verification is needed I wonder if this should be a method at all. Maybe just a function is enough. The main reason for a method is incase users need to configure the underlying client for nextwork requests, but this will not make any, right?

idtoken/validate.go
```diff
@@ -122,6 +122,36 @@ func Validate(ctx context.Context, idToken string, audience string) (*Payload, e
...
+// GetPayload parses the given token and returns its payload.
+//
+// Warning: This function does not validate the token prior to parsing it.
+//
+// GetPayload is primarily meant to be used to inspect a token's payload. This is
+// useful when validation fails and the payload needs to be inspected.
+//
+// Note: A successful Validate() invocation with the same token will return an
+// identical payload.
+func GetPayload(ctx context.Context, idToken string) (*Payload, error) {
```
codyoss, line 151: If the above comment is true I don't think this needs to take a context. Also, maybe `ParsePayload` is a better name as that is what the docs says it does?

idtoken/validate_test.go
```diff
@@ -231,6 +231,42 @@ func TestValidateES256(t *testing.T) {
...
+			ctx := context.Background()
+			v := &Validator{}
+			payload, err := v.GetPayload(ctx, tt.token)
+			gotErr := err != nil
+			if gotErr != tt.wantErr {
+				t.Errorf("GetPayload(ctx, %q) got error %v, wantErr = %v", tt.token, err, tt.wantErr)
+			}
+			if tt.wantPayloadAudience != "" {
+				if payload == nil || payload.Audience != tt.wantPayloadAudience {
+					t.Errorf("GetPayload(ctx, %q) got payload %+v, want payload with audience = %v", tt.token, payload, tt.wantPayloadAudience)
```
noahdietz, line 263:
```suggestion
					t.Errorf("GetPayload(ctx, %q) got payload %+v, want payload with audience = %q", tt.token, payload, tt.wantPayloadAudience)
```

idtoken/validate_test.go
```diff
@@ -231,6 +231,42 @@ func TestValidateES256(t *testing.T) {
 	}
 }
 
+func TestGetPayload(t *testing.T) {
+	idToken, _ := createRS256JWT(t)
+	tests := []struct {
+		name                string
+		token               string
+		wantPayloadAudience string
+		wantPayload         *Payload
```
noahdietz, line 240:
```suggestion
		wantPayloadAudience string
```
`wantPayload` isn't used anywhere in the test or cases.

idtoken/validate.go
noahdietz: +1

idtoken/validate_test.go
eytankidron: Done.

noahdietz (approved):

noahdietz:
LGTM @codyoss please stamp and I will get this into the next release (hopefully today!)

codyoss (approved):

PR #2163, opened 2023-09-15
feat(internal): add some feature flags for new auth libs

codyoss:
We will use the internaloption to start to enable new auth lib for a small number of clients in the future. The envvar will be undocumented and used exclusively in our own testing environments. We will enable it on all of our repos as the first part of functional testing that all existing integration tests continue to work as normal when routing through the new auth layers.

### DIFF
--- a/internal/settings.go
+++ b/internal/settings.go
@@ -10,7 +10,7 @@ import (
 	"errors"
 	"net/http"
 	"os"
-	"strings"
+	"strconv"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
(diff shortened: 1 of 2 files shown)


internal/settings.go
```diff
@@ -77,6 +84,13 @@ func (ds *DialSettings) HasCustomAudience() bool {
 	return len(ds.Audiences) > 0
 }
 
+func (ds *DialSettings) IsNewAuthLibraryEnabled() bool {
+	if ds.EnableNewAuthLibrary || strings.ToLower(os.Getenv(newAuthLibEnVar)) == "true" {
```
noahdietz, line 88: Nit: Perhaps `strconv.ParseBool` with `err != nil` treated as false?

noahdietz (approved):

internal/settings.go
codyoss: good idea!

PR #2185, opened 2023-09-25
chore: give admins ownership of everything

noahdietz:


.github/CODEOWNERS
```diff
@@ -4,3 +4,6 @@
 # Give @yoshi-approver owners for all generated files.
 *.json @yoshi-approver
 *-gen.go @yoshi-approver
+
+# Give admins ownership of Renovate config explicitly
+.github/renovate.json @googleapis/yoshi-go-admins
```
codyoss, line 9: This is fine, but should we just have approve right on .json and -gen.go files too?

codyoss (approved):

.github/CODEOWNERS
noahdietz: Yeah that's the right call. Done.

PR #2186, opened 2023-09-25
fix(docs): add an operation polling example

### LINKED ISSUE #2184: docs: add operation polling example to module docs
https://github.com/googleapis/google-api-go-client/issues/2178 is an example of our docs missing a proper example of how to handle an `Operation` returned by an API call. It is not a "live" `Operation` in that it updates it own state, it needs to be explicitly polled.

So let's add an example to the top level `doc.go` with some examples.

noahdietz:
Fixes #2184

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -93,29 +93,29 @@
 // Operation.Done field indicates it is finished. To do this, use the service's
 // Operation client, and a loop, like so:
 //
-//  op, err := myApiClient.CalculateFoo().Do()
-//  if err != nil {
-//		// handle err
-//  }
+//	 op, err := myApiClient.CalculateFoo().Do()
+//	 if err != nil {
+//			// handle err
+//	 }
 //
-//	operationsService = NewOperationsService(myApiClient)
-//	for {
-//		if op.Done {
-//			break
+//		operationsService = NewOperationsService(myApiClient)
+//		for {
+//			if op.Done {
+//				break
+//			}
+//			// not done, sleep then poll again
+//			time.Sleep(1 * time.Second)
+//
+//			op, err := operationsService.Get(op.Name).Do()
+//			if err != nil {
+//				// handle error
+//			}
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -84,4 +84,38 @@
...
+//		// handle err
+//	 }
+//
+//	 operationsService = NewOperationsService(myApiClient)
+//	 for {
+//		if op.Done {
+//			break
+//		}
+//		// not done, sleep then poll again
+//		time.Sleep(1 * time.Second)
```
codyoss, line 107: I wonder if we should use a gax.Backoff here. That is what we would do internally for gapics, right? Or else maybe consider a slightly larger timeout (or document as such.). Thoughts?

doc.go
```diff
@@ -84,4 +84,38 @@
...
 //	}
+//
+// # Polling Operations
+//
+// If an API call returns an Operation, that means it could take some time to
+// complete the work initiated by the API call. Applications that are interested
+// in the end result of the operation they initiated should wait until the
+// Operation.Done field indicates it is finished. To do this, use the service's
+// Operation client, and a loop, like so:
+//
```
quartzmo, line 95: Do we want to show the import for `NewOperationsService`?

noahdietz: It isn't centrally located, [every API has its own](https://pkg.go.dev/google.golang.org/api@v0.142.0/serviceusage/v1#NewOperationsService). I should include a package alias to indicate that.

doc.go
noahdietz: Sure, I think that's a good idea.

noahdietz: Done.

doc.go
noahdietz: Done.

codyoss (approved):

quartzmo (approved):

PR #2225, opened 2023-10-17
feat(transport): Log DirectPath misconfiguration

mohanli-ml:
We met issues where customers wanted to use DirectPath, but DirectPath was misconfigured and it took 2 weeks for debugging. So we want to add WARNING logs if DirectPath flag/option is set but DirectPath is not used. Issue: https://github.com/googleapis/sdk-platform-java/issues/2164.

Java PR: https://github.com/googleapis/sdk-platform-java/pull/2105.

Note that the log is added when a channel is created because we need to check the OAuth token, which means it will be logged every channel. This is a bad behavior if customer has a large channel pool size. Ideally I want to use something like LOG_EVERY_N_SEC, and I found `rate.Sometimes` can do this. Cody, do you have suggestions here?

@codyoss @frankyn


transport/grpc/dial.go
```diff
@@ -296,6 +297,27 @@ func checkDirectPathEndPoint(endpoint string) bool {
...
 }
 
+func logDirectPathMisconfig(ts oauth2.TokenSource, o *internal.DialSettings) {
+	isDirectPathOptionSet := o.EnableDirectPath
+	isDirectPathXdsEnvSet := strings.EqualFold(os.Getenv(enableDirectPathXds), "true")
+	isDirectPathXdsOptionSet := o.EnableDirectPathXds
+	if isDirectPathXdsEnvSet || isDirectPathXdsOptionSet {
+		// Case 1: does not enable DirectPath
+		if !isDirectPathOptionSet {
+			log.Print("WARNING: DirectPath is misconfigured. Please set the attemptDirectPath option along with the attemptDirectPathXds option.")
```
frankyn, line 307: The name of options are named differently: `EnableDirectPath` and `EnableDirectPathXds`

frankyn, line 307: I'm not familiar enough to warnings in Go but is this how warnings are logged?

codyoss:
If we are going to start logging about this I think we need a nice section in the package docs on how to configure this all correctly. Also logging is not super ideal today because the way this is being done a user could shut off all logging that uses the standard library or keep it all on. Let me think about this. Are we sure we don't want to return an error for misconfiguration? What state will there app be in if this is not working correctly?

transport/grpc/dial.go
mohanli-ml: I did not find a `log.Warn()` API in https://pkg.go.dev/log. @codyoss Hi Cody, could you suggest?

transport/grpc/dial.go
```diff
@@ -296,6 +297,27 @@ func checkDirectPathEndPoint(endpoint string) bool {
...
+		} else {
+			// Case 2: credential is not correctly set
+			if !isTokenSourceDirectPathCompatible(ts, o) {
+				log.Print("WARNING: DirectPath is misconfigured. Please make sure the token source is fetched from GCE metadata server and the default service account is used")
```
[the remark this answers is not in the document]
mohanli-ml: GCS client should set the `AllowNonDefaultServiceAccount` option to true, https://github.com/googleapis/google-api-go-client/blob/main/transport/grpc/dial.go#L269, and customers do not need to worry about this.

mohanli-ml:
> If we are going to start logging about this I think we need a nice section in the package docs on how to configure this all correctly. 

I am not sure about what are the "package docs" you pointed to. My understanding is that each DirectPath service (GCS, CBT, CS) will publish its own doc to its customers.

> Are we sure we don't want to return an error for misconfiguration? What state will there app be in if this is not working correctly?

The app will use CloudPath if DirectPath misconfiguration happens. The purpose here is to give a hint to the customers who want to use DirectPath but traffic is actually CloudPath.

quartzmo (changes requested):

transport/grpc/dial.go
mohanli-ml: Yes you are right, thanks for the clarification.

mohanli-ml: Changed `log.Print` to `log.Println`. Thanks!

transport/grpc/dial.go
```diff
@@ -296,6 +297,24 @@ func checkDirectPathEndPoint(endpoint string) bool {
 	return true
 }
 
+func logDirectPathMisconfig(endpoint string, ts oauth2.TokenSource, o *internal.DialSettings) {
+	if isDirectPathXdsUsed(o) {
```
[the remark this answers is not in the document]
mohanli-ml: Thanks for bringing the concern! Please see my comment in the issue, https://github.com/googleapis/sdk-platform-java/issues/2164#issuecomment-1765423112.

frankyn (approved):
LGTM, thanks @mohanli-ml

frankyn:
We are pending approval from @quartzmo

mohanli-ml:
> We are pending approval from @quartzmo

Can I get some help about how to limit the log? Is `rate.Sometimes` the right thing to use?
```
Note that the log is added when a channel is created because we need to check the OAuth token, which means it will be logged every channel. This is a bad behavior if customer has a large channel pool size. Ideally I want to use something like LOG_EVERY_N_SEC, and I found rate.Sometimes can do this.
```

transport/grpc/dial.go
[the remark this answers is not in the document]
quartzmo: Thanks for the clarification!

quartzmo (approved):

quartzmo:
> Can I get some help about how to limit the log? Is rate.Sometimes the right thing to use?

Yes, I believe that is a good solution. I have not yet personally used it in production however.

mohanli-ml:
I added the rate.Sometimes to limit the log. PTAL. Thanks!

frankyn:
@mohanli-ml IIUC you have to do this because Go doesn't have a similar option to what you did with Java in the constructor? I would think you could check this once without having to do it for every channel instantiation.

mohanli-ml:
> @mohanli-ml IIUC you have to do this because Go doesn't have a similar option to what you did with Java in the constructor? I would think you could check this once without having to do it for every channel instantiation.

The credential code for Java and Go is a bit different. In Java, the [credential](https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L512) is passed to the builder, so the builder can do the check. In Go the credential is created [during channel construction](https://github.com/googleapis/google-api-go-client/blob/main/transport/grpc/dial.go#L142), so builder does not have this information.

Anyways the code will not do the check for every channel instantiation, it is limited to every second.

transport/grpc/dial.go
```diff
@@ -296,6 +304,24 @@ func checkDirectPathEndPoint(endpoint string) bool {
 	return true
 }
 
+func logDirectPathMisconfig(endpoint string, ts oauth2.TokenSource, o *internal.DialSettings) {
```
frankyn, line 307: Please add tests for this function in case changes are made to this code there's expectation on what's expected for this code.

mohanli-ml: Added. PTAL.

frankyn (approved):
Thanks @mohanli-ml one more approval from @quartzmo and let's ship it.

quartzmo (approved):

mohanli-ml:
I do not have the access to merge. @quartzmo Chris, can you help? Thanks!

noahdietz:
> I do not have the access to merge. @quartzmo Chris, can you help? Thanks!

I've added the `automerge` label.

noahdietz:
FYI @mohanli-ml this will be in v0.150.0 https://github.com/googleapis/google-api-go-client/pull/2244

PR #2264, opened 2023-11-16
feat: expose experimental universe-related options

shollyman:
This PR adds the first step in plumbing a universe option into clients
by first exposing it as part of the option package.

It exposes a `WithUniverse(string)` as a public option, and `WithDefaultUniverse(string)`
as an internal option that can be used by generated clients.  It also augments existing
tests.

It does not yet include any behavioral changes.


option/internaloption/internaloption.go
```diff
@@ -126,6 +126,19 @@ func (w withDefaultScopes) Apply(o *internal.DialSettings) {
 	copy(o.DefaultScopes, w)
 }
 
+// WithDefaultUniverseDomain returns a ClientOption that sets the default universe domain.
+//
+// It should only be used internally by generated clients.
```
quartzmo, line 131: Can you add a doc line below this one:

```
//
// 
// This is similar to WithUniverseDomain, but allows us to determine whether the user has overridden the default universe domain.

quartzmo: I saw this comment on `WithDefaultEndpoint`, it seems useful.

quartzmo (approved):

option/internaloption/internaloption.go
shollyman: Done.

PR #2284, opened 2023-12-01
feat: add universe resolution logic

shollyman:
This PR provides utility methods for universe manipulations, and provides an alternative `getEndpointAndUniverse` to the existing `getUniverse`.  A followup PR will replace usages of getEndpoint with this new method, at which point the resolved universe can be propagated further up the stack, as this will eventually need to be exposed in clients.

TODO: testing for all the various endpoint / universe combinations.

### DIFF
--- a/internal/cba.go
+++ b/internal/cba.go
@@ -138,53 +138,6 @@ func isClientCertificateEnabled() bool {
 	return strings.ToLower(useClientCert) == "true"
 }
 
-var universePatternToken = "%%UNIVERSE%%"
-
-// getEndpointAndUniverse returns the endpoint for the service as well as the universe
-// domain, taking in to account the various overrides the user may have provided.
-//
-// This method will also select a default endpoint based on MTLS settings, controlled by
-// the GOOGLE_API_USE_MTLS_ENDPOINT environment variable.
-//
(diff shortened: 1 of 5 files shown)


internal/cba.go
```diff
@@ -138,6 +138,53 @@ func isClientCertificateEnabled() bool {
...
+	// parameterize the default endpoints with the default universe.
+	defUniverse := getDefaultUniverse(settings)
+	defEndpoint := strings.Replace(settings.DefaultEndpoint, defUniverse, universePatternToken, 1)
+	defMTLSEndpoint := strings.Replace(settings.DefaultMTLSEndpoint, defUniverse, universePatternToken, 1)
+
+	universe := getUniverse(settings)
+	if settings.Endpoint == "" {
+		mtlsMode := getMTLSMode()
+		if mtlsMode == mTLSModeAlways || (clientCertSource != nil && mtlsMode == mTLSModeAuto) {
+			return mergeDefaultEndpointUniverse(defMTLSEndpoint, universe), universe, nil
```
quartzmo, line 163: If universe domain other than default is configured here, the exception should be "mTLS is not supported in any universe other than googleapis.com".

shollyman: good catch, updated implementation.

PR #2296, opened 2023-12-08
feat: add universe domain support

quartzmo:
Alternative implementation of #2284.


internal/settings.go
```diff
@@ -19,14 +19,16 @@ import (
...
+	newAuthLibEnVar       = "GOOGLE_API_GO_EXPERIMENTAL_USE_NEW_AUTH_LIB"
+	universeDomainDefault = "googleapis.com"
 )
 
 // DialSettings holds information needed to establish a connection with a
 // Google API service.
 type DialSettings struct {
 	Endpoint                      string
 	DefaultEndpoint               string
+	DefaultEndpointTemplate       string
```
shollyman, line 31: Until such time as service configs provide the template format, I think this should be derived rather than a property of DialSettings.  Or do you envision the generators computing and supplying this at codegen time?

quartzmo: Yes this is an internaloption to be set by the generator

internal/cba.go
```diff
@@ -155,6 +172,10 @@ func getEndpoint(settings *DialSettings, clientCertSource cert.Source) (string,
 	if settings.Endpoint == "" {
 		mtlsMode := getMTLSMode()
 		if mtlsMode == mTLSModeAlways || (clientCertSource != nil && mtlsMode == mTLSModeAuto) {
+			// TODO(chrisdsmith): Closes: CL-R10 (remove this note before publication)
+			if settings.UniverseDomainNotGDU() {
+				return "", errors.New("mTLS is not supported in any universe other than googleapis.com")
```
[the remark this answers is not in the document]
quartzmo: Good call, will export a known error!

internal/settings.go
quartzmo: Also, I found using `mergeEndpoints` to be a bit buggy when using it to replace the universe domain.

internal/cba.go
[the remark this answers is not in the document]
quartzmo: Done, thanks!

impersonate/user.go
```diff
@@ -18,6 +18,8 @@ import (
 	"golang.org/x/oauth2"
 )
 
+// user provides an auth flow for domain-wide delegation, setting
+// CredentialsConfig.Subject to be the  the impersonated user.
```
shollyman, line 22: comment nit:  redundant "the  the"

internal/cba.go
```diff
@@ -94,6 +110,11 @@ func getTransportConfig(settings *DialSettings) (*transportConfig, error) {
 	if !shouldUseS2A(clientCertSource, settings) {
 		return &defaultTransportConfig, nil
 	}
+	if settings.UniverseDomainNotGDU() {
+		return &transportConfig{
```
shollyman, line 114: if you're returning an error, should you build a transport config at all or just return nil?

internal/cba_test.go
```diff
@@ -355,3 +361,141 @@ func setupTest() func() {
...
+			}
+		} else {
+			if tc.wantEndpoint != endpoint {
+				t.Errorf("%s: want endpoint: [%s], got [%s]", tc.name, tc.wantEndpoint, endpoint)
+			}
+		}
+	}
+}
+
+func TestGetGRPCTransportConfigAndEndpoint_UniverseDomain(t *testing.T) {
```
shollyman, line 434: should we have a couple test cases where user effectively overrides via DialSettings.Endpoint?  e.g. UniverseDomain + Endpoint, etc?

internal/creds_test.go
```diff
@@ -120,6 +120,34 @@ func TestJWTWithScope(t *testing.T) {
 	}
 }
 
+func TestJWTWithScopeAndUniverseDomain(t *testing.T) {
```
shollyman, line 123: should we test invalid json does yield an error?  We're really just propagating json.Unmarshal's error up so it's not a particularly useful case.

shollyman (commented):
Left a few minor comments, but overall looks like it's in good shape.

internal/cba.go
quartzmo: I'll change this to `nil`, along with the early return above it from which I copied this code. Not sure of the reason for it. I was feeling a bit risk-adverse to changing it, but static inspection shows the first return value is not used by callers when err is non-nil. (As you would expect!)

Thanks!

internal/creds_test.go
quartzmo: Hmm, what do you suggest doing instead?

internal/cba_test.go
quartzmo: Added, thanks!

shollyman (approved):

impersonate/impersonate.go
```diff
@@ -8,20 +8,25 @@ import (
...
 	"google.golang.org/api/option/internaloption"
 	htransport "google.golang.org/api/transport/http"
 )
 
 var (
-	iamCredentailsEndpoint = "https://iamcredentials.googleapis.com"
-	oauth2Endpoint         = "https://oauth2.googleapis.com"
+	iamCredentailsEndpoint                      = "https://iamcredentials.googleapis.com"
+	oauth2Endpoint                              = "https://oauth2.googleapis.com"
+	ErrUniverseNotSupportedDomainWideDelegation = errors.New(
```
codyoss, line 27: I don't think this needs to be a public error. There is no action for the user to take programmatically if this error occurs.

codyoss: Same with other errors

internal/cba.go
```diff
@@ -67,6 +74,15 @@ func getClientCertificateSourceAndEndpoint(settings *DialSettings) (cert.Source,
 	if err != nil {
 		return nil, "", err
 	}
+	// TODO(chrisdsmith): Use this composed endpoint everywhere to replace DialSettings.DefaultEndpoint
```
codyoss, line 77: nit: instead of tagging these as todo to you can you file issues for this and link to the issue number

internal/settings.go
```diff
@@ -161,3 +162,16 @@ func (ds *DialSettings) Validate() error {
...
+// UniverseDomain returns the default service domain for a given Cloud universe.
+// The default value is "googleapis.com".
+func (ds *DialSettings) GetUniverseDomain() string {
+	if ds.UniverseDomain == "" {
+		return universeDomainDefault
+	}
+	return ds.UniverseDomain
+}
+
+func (ds *DialSettings) UniverseDomainNotGDU() bool {
```
codyoss, line 175: nit: consider starting with `Is`. And do we want the negation in text or code.

impersonate/impersonate.go
shollyman: That's reasonable feedback.  If we end up using the error in multiple places it might make sense to keep it but leave it unexported.

internal/settings.go
quartzmo: Done, thanks!

impersonate/impersonate.go
quartzmo: Fixed, thanks! Also improved the impersonate error unit tests.

internal/cba.go
quartzmo: Done, thanks!

codyoss (approved):

PR #2313, opened 2023-12-18
feat(option/internaloption): add WithDefaultEndpointTemplate

quartzmo:
* Add DefaultEndpointTemplate to internal/settings.go
* Deprecate internaloption.WithDefaultEndpoint

refs: #2264


option/internaloption/internaloption.go
```diff
@@ -22,10 +22,29 @@ func (o defaultEndpointOption) Apply(settings *internal.DialSettings) {
 // It should only be used internally by generated clients.
 //
 // This is similar to WithEndpoint, but allows us to determine whether the user has overridden the default endpoint.
+//
+// Deprecated: WithDefaultEndpoint does not support setting the universe domain.
```
codyoss, line 26: Any reason to not just repurpose this option. I think we could since it is internal?

quartzmo: I actually plan to use both options simultaneously during a transitional period as universe domain support is incrementally expanded.

codyoss: could you do the same thing with one and just check for a %s? I think either way is fine as long as we plan to fully remove the old one eventually

codyoss (approved):

option/internaloption/internaloption.go
quartzmo: Definitely plan to remove the old one.

Issue #2321, opened 2023-12-23
Memory leak in v0.154.0


songjiaxun:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Is this a client library issue or a product issue? We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [Support page](https://cloud.google.com/support) to reach the most relevant engineers.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - Programming language: golang
  - OS: `gcr.io/distroless/base-debian11` image
  - Language runtime version: debian11
  - Package version: [v0.154.0](https://github.com/googleapis/google-api-go-client/releases/tag/v0.154.0)

#### Steps to reproduce

In the [GKE GCS FUSE CSI driver](https://github.com/GoogleCloudPlatform/gcs-fuse-csi-driver/tree/main) project, the `google.golang.org/api` package is used to communicate with the GCS storage API. We found that the feature [feat(transport): add OpenTelemetry context propagation](https://github.com/googleapis/google-api-go-client/pull/2127) might introduce a memory leak.

In our project, we are using `go.opentelemetry.io/otel v1.21.0`, which is the same version [this repo is using](https://github.com/googleapis/google-api-go-client/blob/main/go.mod#L32C2-L32C34).

In our stress test, we found that the `otelgrpc.newConfig` for some reason cannot be cleaned up from the heap. Please see the heap profiling result:

![Bm3NwNiaSiquXso](https://github.com/googleapis/google-api-go-client/assets/31323106/dc9ae71f-db2a-43ff-89d0-bd6e217ae0bf)

As a result, our application memory consumption grows unlimitedly over the time:

![B3VcQdXrEaBoHvF](https://github.com/googleapis/google-api-go-client/assets/31323106/4a7e43f6-b52a-4682-bbd9-e78b3fe8204d)

After we reverted the `google.golang.org/api` package version back to `v0.153.0`, the memory usage went back to normal:

![4WKdJWnQKzVTaJE](https://github.com/googleapis/google-api-go-client/assets/31323106/4734f4e0-b3b3-4771-9b7e-5026f4e29575)

I will DM you for the detailed reproducing and profiling steps.

Making sure to follow these steps will guarantee the quickest resolution possible.

[...]

PR #2329, opened 2024-01-02
fix(transport): fix memory leak in grpc.Dial

quartzmo:
* Memoize otelgrpc.NewClientHandler for reuse

fixes: #2321

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -14,6 +14,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
@@ -27,6 +28,7 @@ import (
 	grpcgoogle "google.golang.org/grpc/credentials/google"
 	grpcinsecure "google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/credentials/oauth"
+	"google.golang.org/grpc/stats"
 
 	// Install grpclb, which is required for direct path.
 	_ "google.golang.org/grpc/balancer/grpclb"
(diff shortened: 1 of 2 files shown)


transport/grpc/dial.go
```diff
@@ -47,6 +49,26 @@ var logRateLimiter = rate.Sometimes{Interval: 1 * time.Second}
...
 
+// otelGRPCClientHandler is a singleton otelgrpc clientHandler to be used across
+// all dial connections to avoid the memory leak documented in
+// https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4226
+//
+// TODO: If 4226 has been fixed in opentelemetry-go-contrib, replace this
+// singleton with inline usage for simplicity.
+var (
+	otelGRPCClientHandler stats.Handler
+	once                  sync.Once
```
codyoss, line 60: nit: just like mutexes with once it is considered better to put the field above the value it protects. Also since these are globals consider a name for the once that aligns it better to the thing it is protecting.

transport/grpc/dial.go
```diff
@@ -47,6 +49,26 @@ var logRateLimiter = rate.Sometimes{Interval: 1 * time.Second}
...
+)
+
+// getOtelGRPCClientHandler returns singleton otelGRPCClientHandler for reuse
+// across all dial connections.
+func getOtelGRPCClientHandler() stats.Handler {
```
codyoss, line 65: nit: remove the word get

codyoss (changes requested):

transport/grpc/dial.go
quartzmo: Done, thanks!

transport/grpc/dial.go
quartzmo: Done, thanks!

codyoss (approved):

PR #2335, opened 2024-01-05
feat(google-api-go-generator): add universe domain support

quartzmo:


google-api-go-generator/gen.go
```diff
@@ -498,6 +498,42 @@ func (a *API) apiBaseURL() string {
 	return resolveRelative(base, rel)
 }
 
+// apiBaseURLTemplate returns the value returned by apiBaseURL with the
+// following changes: 1) The hostname (excluding subdomains) is replaced with %s
+// for universe domain substitution. 2) If the value does not have a scheme,
+// https:// is prepended.
+func (a *API) apiBaseURLTemplate() (string, error) {
+	baseURL := a.apiBaseURL()
+	return replaceDomain(baseURL, "%s")
```
shollyman, line 507: Should we use better sentinel value for the template than "%s"?  My worry is that might be ambigous.

Have you seen any discussions for how APIs might produce the template form directly (via a field in the discovery doc or the API yaml configs)?  This is a reason I didn't consider the template option initially is this is a resolution issue, and so we're simply splitting the resolution logic up between here and elsewhere (e.g. google.golang.org/api/internal).

google-api-go-generator/gen.go
```diff
@@ -498,6 +498,42 @@ func (a *API) apiBaseURL() string {
...
+}
+
+// replaceDomain detects the domain (excluding subdomains) in the provided
+// baseURL and replaces it with newDomain. If baseURL is empty, empty is
+// returned. If baseURL is an invalid URL, the error from url.Parse is returned.
+// If baseURL is valid but does not have a scheme, https:// is prepended to the
+// return value.
+func replaceDomain(baseURL, newDomain string) (string, error) {
```
shollyman, line 515: Should we perform the replacement if the GDU (googleapis.com) isn't in the baseURL?

I'm considering two cases:
* users who might be using the generator with non-google APIs
* alphabet/google services that don't support universes

It might be better to limit templating to "trusted" domains, which for our case is simple googleapis.com at present.

google-api-go-generator/gen.go
quartzmo: > Should we use better sentinel value for the template than "%s"?

I'm totally open to this. I can update the PR with whatever string seems most common across languages.

> we're simply splitting the resolution logic up between here and elsewhere

I considered this as well. My conclusion was that it's important to make the replacement mechanism explicit to users by configuring the template and the default universe domain in the client setup code, even at the cost of duplicating the messy work of creating the template string. By moving this upstream to the generator, things are going to be clear(er) for users who inspect the source of their clients. Dynamically changing the endpoint passed in `WithDefaultEndpoint` after client init seems much harder for users to comprehend.

google-api-go-generator/gen.go
quartzmo: Good point. I don't know. @viacheslav-rostovtsev any comment on this?

google-api-go-generator/gen.go
quartzmo: I changed the sentinel value from `%s` to `UNIVERSE_DOMAIN`, thanks!

codyoss (approved):
LGTM. Is mTLS in scope of universe stuff? Or is that later?

quartzmo:
@codyoss My understanding was that the default mTLS endpoint should be left as-is for now, but I'll double-check.

quartzmo:
@codyoss: Per @viacheslav-rostovtsev, "currently configuring MTLS with TPC should return an error", so the`mtlsBasePath` can remain static.

google-api-go-generator/gen.go
quartzmo: Per @viacheslav-rostovtsev:  if the generator receives a default endpoint like `ths-prod.googleplex.com/_ah/api/tshealth/v1/`, the default endpoint template should be the unchanged value, `ths-prod.googleplex.com/_ah/api/tshealth/v1/`. (And not `https://ths-prod.UNIVERSE_DOMAIN/_ah/api/tshealth/v1/`.) Only `googleapis.com` should be replaced.

quartzmo: Updated to only replace GDU (googleapis.com) in the baseURL, thanks!

shollyman (approved):
Thanks for getting this together!

PR #2355, opened 2024-01-17
feat(transport): add universe domain support

quartzmo:


quartzmo:
Extracted from #2296.

transport/http/dial.go
```diff
@@ -101,6 +109,15 @@ func newTransport(ctx context.Context, base http.RoundTripper, settings *interna
 	return trans, nil
 }
 
+func errUniverseNotMatch(settingsUD, credsUD string) error {
+	return fmt.Errorf(
+		"the configured universe domain (%s) does not match the universe "+
```
codyoss, line 114: nit: instead of `(%s)` just use `%q`

transport/http/dial.go
```diff
@@ -88,6 +89,13 @@ func newTransport(ctx context.Context, base http.RoundTripper, settings *interna
 		if err != nil {
 			return nil, err
 		}
+		credsUniverseDomain, err := creds.GetUniverseDomain()
+		if err != nil {
+			return nil, err
+		}
+		if settings.GetUniverseDomain() != credsUniverseDomain {
```
codyoss, line 96: Won't this always happen if someone sets the value with WithUniverseDomain and creds returns the default?

codyoss (changes requested):

transport/http/dial.go
quartzmo: Correct.

> The verification can be done by comparing the universe_domain from the end-user configuration to the universe_domain from the credentials. If they match, the credentials can be used, if not the client libraries must raise an error instead of making a service call to the user-specified universe.

transport/http/dial.go
quartzmo: Fixed, thanks!

transport/http/dial.go
codyoss: What is the point of the option in that case? They will never match unless it is set in the creds. Because we return a default value: https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.16.0:google/default.go;l=92

shollyman: I'm not seeing this creds method directly, do we have cases where creds have no universe embedded?  Do we need to check for a non-empty universe from creds?

transport/http/dial_test.go
```diff
@@ -37,3 +38,14 @@ func TestNewClient(t *testing.T) {
 		t.Fatalf("got %s, want: %s", got, want)
 	}
 }
+
+func TestNewClient_UniverseDomain(t *testing.T) {
```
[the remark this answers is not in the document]
quartzmo: Yes, but the creds (unseen in this test) do not have the same universe domain "example.com". They have the GDU.

quartzmo:
> I'm not seeing this creds method directly, do we have cases where creds have no universe embedded? Do we need to check for a non-empty universe from creds?

See behavior of creds at https://github.com/golang/oauth2/blob/master/google/default.go#L91-L92

transport/http/dial.go
quartzmo: > They will never match unless it is set in the creds. 

Here's more spec:

> Credentials are universe-specific. Whether a user provides the credentials or they are received via ADC, the credentials object (or equivalent) is expected to have the universe_domain property. The Client Libraries must read this property in order to verify that the credentials are valid for the user-supplied universe domain.

quartzmo: @shollyman 

> I'm not seeing this creds method directly, do we have cases where creds have no universe embedded? Do we need to check for a non-empty universe from creds?

See behavior of creds at https://github.com/golang/oauth2/blob/master/google/default.go#L91-L92

transport/http/dial_test.go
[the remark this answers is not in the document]
quartzmo: I'm going to update the test with explicit credentials.

transport/http/dial.go
shollyman: Thanks for the reference.  I think my largest concern is related to non-cloud resolution (e.g. a service not in GDU currently).  With the current implementation, both settings and creds yield up the GDU as the default value so I don't _think_ it's an issue, but it is something for future consideration.

transport/http/dial_test.go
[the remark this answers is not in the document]
shollyman: Yeah, either that or using the test name to signal that this is a credentials test.

transport/http/dial.go
quartzmo: @shollyman This would be good to bring up with the spec authors. Afaik, the Go implementation adheres to the spec.

transport/http/dial_test.go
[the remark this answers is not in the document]
quartzmo: @shollyman Can you suggest a better test name?

transport/http/dial.go
codyoss: After looking at this with fresh eyes and thinking though it I think this is fine here, but really feels like it should be in the oauth2 layer. If we set the the override that is plumbed through to the oauth2 layer and is the value used by the credential I think? So it feels like this should never err perhaps? If that is the case it should be a test and not a in code validation. But I y'all are on this more than me. Approving.

codyoss (approved):

transport/http/dial_test.go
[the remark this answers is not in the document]
shollyman: With the explicit creds changes it's clearer, but something like TestNewClient_MismatchedUniverseCreds might have made this cleaer.

shollyman (approved):

transport/http/dial.go
quartzmo: The `WithUniverseDomain` override is in fact NOT plumbed through to the oauth2 layer, it is only used to construct the endpoint in the client, pending a number of error condition checks like the one discussed here. The `creds` instance here needs to have the same universe domain value as is provided to `settings`, but only `settings` gets it via the `WithUniverseDomain` override. The `creds` instance gets it either from credentials JSON or from oauth2's `google.CredentialsParams`. If the two separately-sourced values do not match, we need to return an error here.

transport/http/dial_test.go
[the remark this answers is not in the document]
quartzmo: Updated, thanks!

PR #2356, opened 2024-01-17
docs(option): update WithDefaultEndpointTemplate docs

quartzmo:
Follow-up to #2335.

### DIFF
--- a/option/internaloption/internaloption.go
+++ b/option/internaloption/internaloption.go
@@ -38,7 +38,10 @@ func (o defaultEndpointTemplateOption) Apply(settings *internal.DialSettings) {
 
 // WithDefaultEndpointTemplate provides a template for creating the endpoint
 // using a universe domain. See also WithDefaultUniverseDomain and
-// option.WithUniverseDomain.
+// option.WithUniverseDomain. The placeholder UNIVERSE_DOMAIN should be used
+// instead of a concrete universe domain such as "googleapis.com".
+//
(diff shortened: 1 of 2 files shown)


option/internaloption/internaloption.go
```diff
@@ -163,6 +166,10 @@ func (w withDefaultUniverseDomain) Apply(o *internal.DialSettings) {
 
 // EnableJwtWithScope returns a ClientOption that specifies if scope can be used
 // with self-signed JWT.
+//
+// EnableJwtWithScope is ignored when option.WithUniverseDomain is set
+// to a non-GDU universe domain. For non-GDU universe domains, token exchange is
```
codyoss, line 171: What is a GDU, would a user know this term?

quartzmo: Good call, I'll expand it.

quartzmo: Actually, this is internaloption, is it ok here? Or do you think I should spell it out for maintainers as well.

codyoss: I think it is still worth while as I am not sure what it stand for

quartzmo: "Google Default Universe" fwiw

codyoss (approved):

option/internaloption/internaloption.go
quartzmo: Fixed, thanks!

PR #2368, opened 2024-01-23
fix(transport): not enable s2a when there is endpoint override

xmenxk:
don't enable s2a if an override endpoint is set, since we don't know if the endpoint is mtls-capable or not

### DIFF
--- a/internal/cba.go
+++ b/internal/cba.go
@@ -111,11 +111,6 @@ func getTransportConfig(settings *DialSettings) (*transportConfig, error) {
 		return nil, errUniverseNotSupportedMTLS
 	}
 
-	s2aMTLSEndpoint := settings.DefaultMTLSEndpoint
-	// If there is endpoint override, honor it.
-	if settings.Endpoint != "" {
-		s2aMTLSEndpoint = endpoint
-	}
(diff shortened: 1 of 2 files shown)


internal/cba.go
```diff
@@ -111,11 +111,6 @@ func getTransportConfig(settings *DialSettings) (*transportConfig, error) {
 		return nil, errUniverseNotSupportedMTLS
 	}
 
-	s2aMTLSEndpoint := settings.DefaultMTLSEndpoint
-	// If there is endpoint override, honor it.
-	if settings.Endpoint != "" {
```
quartzmo, line 116: Is it a breaking change for any users to remove this?

quartzmo: Also, maybe the scope should be `fix` rather than `feat`? I'm not sure what feature this is adding. Seems like more of a bug fix.

xmenxk: thanks
The purpose of this PR is to not use s2a (new behavior) in cases where Endpoint is set. So the library will continue to follow its default behavior, i.e. TLS to the DefaultEndpoint, which should be safe.

quartzmo (approved):

PR #2375, opened 2024-01-25
fix(transport/grpc): add universe domain verification

quartzmo:


internal/creds.go
```diff
@@ -226,3 +226,12 @@ func baseTransport() *http.Transport {
 		ExpectContinueTimeout: 1 * time.Second,
 	}
 }
+
+func ErrUniverseNotMatch(settingsUD, credsUD string) error {
+	return fmt.Errorf(
+		"the configured universe domain (%q) does not match the universe "+
+			"domain found in the credentials (%q). If you haven't configured "+
+			"WithUniverseDomain explicitly, googleapis.com is the default",
```
noahdietz, line 234: Nit, all other universe domain values are wrapped in double quotes - should `googleapis.com` be as well?

quartzmo: Fixed, thanks!

internal/creds.go
```diff
@@ -226,3 +226,12 @@ func baseTransport() *http.Transport {
 		ExpectContinueTimeout: 1 * time.Second,
 	}
 }
+
+func ErrUniverseNotMatch(settingsUD, credsUD string) error {
```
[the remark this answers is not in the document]
quartzmo: Fixed, thanks!

PR #2376, opened 2024-01-26
fix(transport): relax universe checks

shollyman:
This PR serves to relax universe mismatch checks, as a user specifying credentials via the option.WithTokenSource has provided a form of credential where the universe information is not directly accessible.

In these cases, we no longer perform the universe mismatch checks.

This PR also augments the existing mismatch checks to exercise a more diverse set of client options.


quartzmo (approved):
Thank you!

Issue #2379, opened 2024-01-26
generator: panic when parameter is repeated Any portrayed as "type: object"


noahdietz:
An API with RPC request message `google.protobuf.HttpBody` without a `(google.api.http).body` set will end up with its `repeated google.protobuf.Any extensions` field portrayed in Discovery as a `type: array` with items `type: object` and `additionalProperties` containing `type: any`. The `simpleTypeConvert` function isn't handling `type: object` case so it returns `!ok` and the resulting usage `panic`s, but it would've handled `type: any` had the `additionalProperties` been consulted for that field.

Without considering a change to the Discovery document (which we should probably consult the Discovery generation owner about), our options include:
1. Handling `type: object` in `simpleTypeConvert`, potentially in the same way as `any`, using an `interface{}`
2. Consult the `additionalProperties` for potential a `type` field

See [this internal post](https://yaqs.corp.google.com/eng/q/5229634642981683200#a1) for my detailed debugging process.

noahdietz:
In the specific case, the issue with the API is actually due the fact that a `repeated Any` is used as query parameter. Based on the google.api.http documentation `repeated` non-primitive fields are prohibited from use as query params. So the case that sparked this issue, we would well to call to improve the error messaging.

We also probably still want to handle the `simpleTypeConvert` case handling for this instance too though.

noahdietz:
Fix for handling the "object as query parameter" case been released. This ultimately should've been prevented in API service compilation which is where the real fix will be. Closing this now.

PR #2383, opened 2024-01-28
fix(gen): reject repeated object query params

noahdietz:
Updates #2379


quartzmo:
When do you plan to make this ready to review? What's pending?

noahdietz:
> When do you plan to make this ready to review? What's pending?

I wanted to run a TGP with the patch internally to see if it would catch anything unintended 😅

noahdietz:
Ok, TGP passed with the only broken targets being completely unrelated! So let's get this in.

codyoss (approved):

PR #2385, opened 2024-01-30
fix(transport): skip s2a for now if service has direct path enabled

xmenxk:


internal/cba.go
```diff
@@ -310,6 +310,9 @@ func shouldUseS2A(clientCertSource cert.Source, settings *DialSettings) bool {
 	if settings.HTTPClient != nil {
 		return false
 	}
+	if settings.EnableDirectPath == true || settings.EnableDirectPathXds == true {
+		return false
+	}
 	return true
```
noahdietz, line 316:
```suggestion
	return !settings.EnableDirectPath && !settings.EnableDirectPathXds
```

This would be simpler, if both are false, return true. If either are true, return false.

noahdietz (approved):

PR #2393, opened 2024-02-03
fix(transport): enforce 1s timeout on requests to MDS universe_domain

quartzmo:
* Temporarily return googleapis.com on MDS universe_domain timeout.

closes: googleapis/google-cloud-go#9350


internal/settings.go
```diff
@@ -187,3 +188,32 @@ func (ds *DialSettings) GetUniverseDomain() string {
...
+func GetUniverseDomain(creds *google.Credentials) (string, error) {
+	timer := time.NewTimer(time.Second)
+	defer timer.Stop()
+	var credsUniverseDomain string
+	var err error
+	go func() {
+		credsUniverseDomain, err = creds.GetUniverseDomain()
+	}()
+	<-timer.C
+	if err != nil {
```
codyoss, line 205: This code is racy. You are reading values that are updated in another goroutine without synchronization. I would recommend using channels and a select statement.

internal/settings.go
```diff
@@ -187,3 +188,32 @@ func (ds *DialSettings) GetUniverseDomain() string {
 func (ds *DialSettings) IsUniverseDomainGDU() bool {
 	return ds.GetUniverseDomain() == ds.GetDefaultUniverseDomain()
 }
+
+// GetUniverseDomain returns the default service domain for a given Cloud
+// universe, from google.Credentials, for comparison with the value returned by
+// (*DialSettings).GetUniverseDomain. This wrapper function should be removed
+// to close [TODO(chrisdsmith): issue link here]. See details below.
+func GetUniverseDomain(creds *google.Credentials) (string, error) {
+	timer := time.NewTimer(time.Second)
```
codyoss, line 197: Is one second long enough? Seems short, but I don't know the slo of the metadata service.

codyoss (changes requested):

internal/settings.go
quartzmo: @viacheslav-rostovtsev says 1s is appropriate.

internal/settings.go
quartzmo: Updated, PTAL.

internal/settings.go
```diff
@@ -196,24 +196,44 @@ func (ds *DialSettings) IsUniverseDomainGDU() bool {
...
-		// the endpoint and will timeout. This is the source of issues such as
-		// https://github.com/googleapis/google-cloud-go/issues/9350.
-		// Temporarily (2024-02-02) return the GDU domain. Restore the original
-		// calls to creds.GetUniverseDomain() in grpc/dial.go and http/dial.go
-		// and remove this method to close [TODO(chrisdsmith): issue link here].
-		return universeDomainDefault, nil
-	}
-	return credsUniverseDomain, nil
+
+loop:
```
codyoss, line 211: This can all be simplified a bit. You should be able to use a single select without a loop. The select will wait until one of the channels returns a result.

viacheslav-rostovtsev (commented):
LGTM

viacheslav-rostovtsev (approved):

internal/settings.go
quartzmo: Thanks, updated.

codyoss (approved):

PR #2446, opened 2024-03-01
fix(transport): check TokenSource in obtained credentials

laurynasgadl:
Adjusts universe mismatch conditional to check for TokenSource in credentials obtained after running through all the dial options.

This addresses cases where TokenSource is filled by options other than `WithTokenSource`, such as `WithCredentialsJSON` and `WithCredentialsFile`.

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -177,7 +177,7 @@ func dial(ctx context.Context, insecure bool, o *internal.DialSettings) (*grpc.C
 			if err != nil {
 				return nil, err
 			}
-			if o.TokenSource == nil {
+			if creds.TokenSource == nil {
 				// We only validate non-tokensource creds, as TokenSource-based credentials
 				// don't propagate universe.
 				credsUniverseDomain, err := internal.GetUniverseDomain(creds)
--- a/transport/http/dial.go
+++ b/transport/http/dial.go
@@ -88,7 +88,7 @@ func newTransport(ctx context.Context, base http.RoundTripper, settings *interna
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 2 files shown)


laurynasgadl:
Related to: https://github.com/googleapis/google-api-go-client/pull/2376

This would also fix a new issue we started experiencing, where creating a new PubSub client with service-account credentials on a non-GCE system started throwing:
`pubsub(publisher): Get "http://169.254.169.254/computeMetadata/v1/universe/universe_domain": dial tcp 169.254.169.254:80: connect: no route to host`

after upgrading these dependencies:
```
cloud.google.com/go/pubsub v1.33.0 => v1.36.1
google.golang.org/api v0.126.0 => v0.165.0
google.golang.org/grpc v1.56.3 => v1.61.1
```

quartzmo:
@laurynasgadl Thank you for this pull request. Unfortunately, we can't merge it, because it would disable the universe domain mismatch check on auth flows where it is needed.

Can you please open a  [new google-cloud-go bug issue](https://github.com/googleapis/google-cloud-go/issues/new?assignees=&labels=triage+me&projects=&template=bug_report.md&title=packagename%3A+short+description+of+bug) (please use this link!)  for the error mentioned in your [comment above](https://github.com/googleapis/google-api-go-client/pull/2446#issuecomment-1972890058) with as much detail about your environment as you can provide? The request in your error should not occur in a non-GCE environment, so any guesses you might have about why [metadata.OnGCE](https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/metadata.go#L113-L189) is apparently returning `true` in your environment would be very helpful. Thank you so much!

PR #2509, opened 2024-04-05
feat(option): add support for env var GOOGLE_CLOUD_UNIVERSE_DOMAIN

quartzmo:
* Remove unneeded DefaultUniverseDomain usages.


internal/settings.go
```diff
@@ -165,31 +168,37 @@ func (ds *DialSettings) Validate() error {
...
-// with internaloption.WithDefaultUniverseDomain.
+// universe, with the following precedence:
+//
+// 1. A non-empty option.WithUniverseDomain.
+// 2. A non-empty environment variable GOOGLE_CLOUD_UNIVERSE_DOMAIN.
+// 3. The default value "googleapis.com".
 func (ds *DialSettings) GetUniverseDomain() string {
-	if ds.UniverseDomain == "" {
-		return ds.GetDefaultUniverseDomain()
+	ds.udMu.Lock()
```
codyoss, line 184: Instead of updating this value and adding a mutex can we just return the values?

quartzmo: Yes, that should be fine with the new `auth` library, which will only access `GetUniverseDomain` on new client initialization afaik.

quartzmo: Updated, thanks.

internal/settings.go
```diff
@@ -165,31 +165,32 @@ func (ds *DialSettings) Validate() error {
...
-// universe, as configured with internaloption.WithDefaultUniverseDomain.
-// The default value is "googleapis.com".
+// GetDefaultUniverseDomain returns the Google default universe domain
+// ("googleapis.com").
 func (ds *DialSettings) GetDefaultUniverseDomain() string {
-	if ds.DefaultUniverseDomain == "" {
-		return universeDomainDefault
-	}
-	return ds.DefaultUniverseDomain
+	return defaultUniverseDomain
```
codyoss, line 171: nit: I don't think we need this method, do we?

quartzmo: It's an exported method, I didn't want to remove it.

codyoss: its in an internal package. If we need it fine, but we should also feel fine removing it and or making the const public instead

PR #2544, opened 2024-04-25
feat: reduce code size by 29%

codyoss:
- Move more common docs to package level and refer to them
- Make some docs more concise
- Remove some whitespace
- inline some code
- refactor setting common headers to gensupport
- remove all non-user facing code comments

PR #2546, opened 2024-04-25
fix: split large compute file up

codyoss:
This is a followup to #2544. That change reduced the size of the compute client from 9.4M to 6.7. This change now splits the compute client into three files ranging from 2.7-1.8M each. This should allow editors and tooling to better be able to process the client.


google-api-go-generator/gen.go
```diff
@@ -341,7 +347,49 @@ func writeFile(file string, contents []byte) error {
...
+			var err error
+			newB, err = format.Source(b)
+			if err != nil {
+				return err
+			}
+		} else {
+			// determine the new file name
+			base := filepath.Dir(file)
+			fileNum := i + 1
+			name = base + string(filepath.Separator) + pkg + fmt.Sprint(fileNum) + "-gen.go"
```
noahdietz, line 372: Nit: just making use of std lib a bit more
```suggestion
			name = filepath.Join(base, fmt.Sprintf("%s%d-gen.go", pkg, fileNum))
```

google-api-go-generator/gen.go
```diff
@@ -341,7 +347,49 @@ func writeFile(file string, contents []byte) error {
...
+
+			// prepend file header, package, and imports
+			var buf bytes.Buffer
+			// Size the buffer so there is room for the header and contents
+			buf.Grow(len(b) + 500)
```
noahdietz, line 377: Nit: can we move the magic number into a descriptive constant

google-api-go-generator/gen.go
```diff
@@ -874,6 +926,56 @@ func (a *API) GenerateCode() ([]byte, error) {
...
+			panic(err)
+		}
+	}
+
+	pn(`// Copyright %s Google LLC.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Code generated file. DO NOT EDIT.
+`, *copyrightYear)
```
noahdietz, line 952: Nit: Any reason this is a multiline string and other examples are single line per print?
```suggestion
	pn("// Copyright %s Google LLC.", *copyrightYear)
	pn("// Use of this source code is governed by a BSD-style")
	pn("// license that can be found in the LICENSE file.")
	pn("")
	pn("// Code generated file. DO NOT EDIT.")
```

Don't mind either way.

noahdietz (approved):
Really really cool.

google-api-go-generator/gen.go
codyoss: done

PR #2549, opened 2024-04-26
transport/http: use ipv4 explicitly for dialers if ipv6 is not supported

melinath:
This is a workaround for https://github.com/golang/go/issues/25321, and is related to https://github.com/hashicorp/terraform-provider-google/issues/6782

For additional context, see yaqs/5034757202074664960 and yaqs/47302089738551296.


melinath:
The conventionalcommits linter conflicts with https://github.com/googleapis/google-api-go-client/blob/main/CONTRIBUTING.md - which should I follow?

melinath:
I'll assume conventionalcommits for now.

enocom:
Related to https://github.com/hashicorp/terraform-provider-google/issues/6782

transport/http/dial.go
```diff
@@ -266,6 +267,14 @@ func defaultBaseTransport(ctx context.Context, clientCertSource cert.Source, dia
 	if trans == nil {
 		trans = fallbackBaseTransport()
 	}
+	trans.DialContext = func(ctx context.Context, network string, addr string) (net.Conn, error) {
+		// Don't try IPv6 if it's not supported.
+		// https://github.com/golang/go/issues/25321
+		if !nettest.SupportsIPv6() {
```
noahdietz, line 273: I'm a bit hesitant to use a package meant for testing in a production context. Is this what the Go team recommends? Is there any other package we can use that is meant for critical path usage?

noahdietz (commented):
I'd like @codyoss to weigh in here too

noahdietz:
> The conventionalcommits linter conflicts with https://github.com/googleapis/google-api-go-client/blob/main/CONTRIBUTING.md - which should I follow?

I'll fix this, thanks for the shout out.

transport/http/dial.go
codyoss: I think this package would be fine to use as it is meant to test the network and not necessary unit tests. This does feel like code that should happen in terraform though maybe and not here. I think the reason for putting it here is that our transport function is not handling some of the mTLS bits that our NewHTTPClient function does. Is that correct @melinath? If so maybe we can move that logic ad NewHTTPClient calls into NewTransport.

codyoss: > Some of the APIs in that package take testing.T which gave me pause.

I had not noticed that at first glance. That does give me more pause.

codyoss:
I think because of how this is constructed in our new auth library it will not be an issue -- we allow passing a base roundtripper: https://pkg.go.dev/cloud.google.com/go/auth/httptransport#Options

I am less inclined to accept a change here that does this in our base transport. As a workaround you could likely still use NewClient and override http.DefaultTransport as we use that as our base if it is an http.Transport: https://github.com/googleapis/google-api-go-client/blob/4b79cc4e7d85f3a05e5ca99738cdf340b46da243/transport/http/dial.go#L265C11-L265C26

melinath:
Just to clarify, are you suggesting that I should override the global `http.DefaultTransport` variable? That seems like it could have unintended side effects.

This bug impacts all users of this repository, so it would be great to get it fixed here rather than having to do a hacky workaround in every user of this library. (Better yet would be if Go core could fix the issue but they don't seem inclined to do so and I can't figure out how to make the fix.)

That said, if you don't want to fix the bug, would you at least accept a contribution to allow passing in a new default transport rather than requiring an override of a global variable?

melinath:
For some discussion of how modifying the global state of DefaultTransport can be problematic, see https://github.com/hashicorp/go-cleanhttp?tab=readme-ov-file#cleanhttp

melinath:
@noahdietz @codyoss any updates on whether you will consider this contribution? It would be great to fix this issue here instead of downstream users of the libraries!

codyoss:
cc @dneil do you have any thoughts here? If we were going to add something like this is there an easier way to probe for ipv6?

PR #2563, opened 2024-05-01
feat(google-api-go-client): add x-goog-api-version header

quartzmo:


noahdietz (approved):

google-api-go-generator/gen.go
```diff
@@ -2224,6 +2224,10 @@ func (meth *Method) generateCode() {
 		contentType = `"application/json"`
 	}
 	pn(`reqHeaders := gensupport.SetHeaders(c.s.userAgent(), %s, c.header_ )`, contentType)
+	apiVersion := meth.m.APIVersion
```
codyoss, line 2227: nit: Instead of inlining this to every function, could we add a new param to SetHeaders and send this in there?

codyoss (approved):

PR #2572, opened 2024-05-07
[DON'T LAND YET]Allow setting up availabilityDomain in Scheduling object

AmyWrich:
This is for creating internal patch, please don't land.

Allow setting up the availability domain in the scheduling message. So that newly spawned cloud hosts can have the domain set up. The scheduling structure currently includes availability domain in the alpha API but not yet in production. We need to enable the setup of availability domain in launching instance requests to unblock the use of GCP placement groups.

### DIFF
--- a/compute/v1/compute-api.json
+++ b/compute/v1/compute-api.json
@@ -58332,6 +58332,11 @@
           "description": "Specifies whether the instance should be automatically restarted if it is terminated by Compute Engine (not terminated by a user). You can only set the automatic restart option for standard instances. Preemptible instances cannot be automatically restarted. By default, this is set to true so an instance is automatically restarted if it is terminated by Compute Engine.",
           "type": "boolean"
         },
+        "availabilityDomain": {
+          "description": "Specifies the availability domain (AD), which this instance should be scheduled on. The AD belongs to the spread GroupPlacementPolicy resource policy that has been assigned to the instance. Specify a value between 1-max count of availability domains in your GroupPlacementPolicy. See go/placement-policy-extension for more details.",
+          "format": "int32",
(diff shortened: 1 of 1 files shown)


AmyWrich:
@codyoss Do you know how is https://raw.githubusercontent.com/googleapis/google-api-go-client/main/compute/v1/compute-gen.go get generated? is it based on this json file? I am trying to create an internal patch.

codyoss:
It is generated from the public spec of the API: https://www.googleapis.com/discovery/v1/apis/compute/v1/rest
Using the code generator in this repo: https://github.com/googleapis/google-api-go-client/tree/main/google-api-go-generator

AmyWrich:
@codyoss So even I update this compute/v1/compute-api.json, I still wont have the new field in https://raw.githubusercontent.com/AmyWrich/google-api-go-client/7eeb0208dce5e1dcaed5103c1c77b5772a827b83/compute/v1/compute-gen.go?

More context is 
I'm currently using **google.golang.org/api/compute/v1** , which hasn't incorporated this update yet. GCP said will deliver end of June but I need it sooner. is there any way I can update this library locally to have this change?

codyoss:
You would need the server to have the changes incorporated too, so I don't think adding it here to the client would work unless the backend changes have also taken affect.

If the backend changes are active you could always create an authenticated client and make the raw request yourself: https://pkg.go.dev/google.golang.org/api/transport/http#NewClient. You would need to pass similar options that the generated compute client sends into this method for it to work (notably scopes for auth)

AmyWrich:
Yea, maybe http is the easy way to do. Thank you!

PR #2610, opened 2024-05-28
fix(cba): Update credsNewAuth to support oauth2 over mTLS

andyrzhao:
this logic is ported over from "baseCreds" from the same file.


codyoss:
@andyrzhao would you want to up to the latest version of cloud.google.com/go/auth in the PR and test to make sure the patch works with your tests?

andyrzhao:
> @andyrzhao would you want to up to the latest version of cloud.google.com/go/auth in the PR and test to make sure the patch works with your tests?

Yes I verified that this patch works correctly with the following:
replace cloud.google.com/go/auth => github.com/googleapis/google-cloud-go/auth 685784ea84358c15e9214bdecb307d37aa3b6d2f

Can you wait until the go auth change is released before merging in this patch? Thanks!

codyoss:
Created: https://github.com/googleapis/google-api-go-client/pull/2621

codyoss (dismissed):

codyoss (approved):

PR #2612, opened 2024-05-29
chore(gensupport): merge x-goog-api-client vals into a single header

BrennaEpp:


quartzmo (changes requested):
Can you expand `TestSendRequestHeader` (or add a new test) to cover this change?

internal/gensupport/send.go
```diff
@@ -48,8 +48,24 @@ func SendRequest(ctx context.Context, client *http.Client, req *http.Request) (*
 	if ctx != nil {
 		headers := callctx.HeadersFromContext(ctx)
 		for k, vals := range headers {
-			for _, v := range vals {
-				req.Header.Add(k, v)
+			if k != "x-goog-api-client" {
+				for _, v := range vals {
```
tritone, line 52: nit: could you flip this conditional? Seems more readable to have the default case as the "else".

tritone (approved):
Overall looks good; thanks for adding the test.

quartzmo (approved):

shollyman (approved):

tritone (approved):

Issue #2613, opened 2024-05-30
Import error when updating to v0.182.0


CoreFloDev:
Hi,

I got another issue similar to https://github.com/googleapis/google-api-go-client/issues/2543

when I update to v0.182.0, there is:

- github.com/googleapis/gax-go/v2 v2.12.4 // indirect that break with
```
error while importing github.com/googleapis/gax-go/v2/apierror/internal/proto: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/code in multiple modules:
	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 (/home/florent/go/pkg/mod/google.golang.org/genproto@v0.0.0-20200526211855-cb27e3aa2013/googleapis/rpc/code)
	google.golang.org/genproto/googleapis/rpc v0.0.0-20240521202816-d264139d666e (/home/florent/go/pkg/mod/google.golang.org/genproto/googleapis/rpc@v0.0.0-20240521202816-d264139d666e/code)
error while importing github.com/googleapis/gax-go/v2/apierror: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/errdetails in multiple modules:
	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 (/home/florent/go/pkg/mod/google.golang.org/genproto@v0.0.0-20200526211855-cb27e3aa2013/googleapis/rpc/errdetails)
	google.golang.org/genproto/googleapis/rpc v0.0.0-20240521202816-d264139d666e (/home/florent/go/pkg/mod/google.golang.org/genproto/googleapis/rpc@v0.0.0-20240521202816-d264139d666e/errdetails)
```

- google.golang.org/grpc v1.64.0 // indirect also break with
```
error while importing google.golang.org/grpc/status: ambiguous import: found package google.golang.org/genproto/googleapis/rpc/status in multiple modules:
	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013 (/home/florent/go/pkg/mod/google.golang.org/genproto@v0.0.0-20200526211855-cb27e3aa2013/googleapis/rpc/status)
	google.golang.org/genproto/googleapis/rpc v0.0.0-20240521202816-d264139d666e (/home/florent/go/pkg/mod/google.golang.org/genproto/googleapis/rpc@v0.0.0-20240521202816-d264139d666e/status)
```
The previous version v0.181.0 is working fine but not that one, do you get why this is happening?

Thanks!

[...]

Issue #2559, opened 2024-04-30
deps: remove forced dependency on main google-cloud-go module


noahdietz:
We need to force a dependency on the `cloud.google.com/go` main module to ensure that the module is resolved unambiguously. The issue encountered in #2543 and present in #2552 (see below) is odd though because we have other dependencies with newer, unambiguous dependencies on this module. These *should* win out during dep resolution, but they aren't for some reason.

Forcing a direct dep on the module resolves the ambiguity but is not an ideal solution.

```
/root/go/pkg/mod/cloud.google.com/go/auth@v0.3.0/internal/internal.go:31:2: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
	cloud.google.com/go v0.26.0 (/root/go/pkg/mod/cloud.google.com/go@v0.26.0/compute/metadata)
	cloud.google.com/go/compute/metadata v0.3.0 (/root/go/pkg/mod/cloud.google.com/go/compute/metadata@v0.3.0) (compile)
```

quartzmo:
> we have other dependencies with newer, unambiguous dependencies on this module

Can you explicitly document what these are?

noahdietz:
> > we have other dependencies with newer, unambiguous dependencies on this module
> 
> Can you explicitly document what these are?

At `HEAD` before forcing the dep:
```
$ go mod graph | grep ' cloud.google.com/go@'

cloud.google.com/go/compute@v1.20.1 cloud.google.com/go@v0.110.2
cloud.google.com/go/compute@v1.24.0 cloud.google.com/go@v0.112.0
google.golang.org/genproto@v0.0.0-20240227224415-6ceb2ff114de cloud.google.com/go@v0.112.0
google.golang.org/genproto@v0.0.0-20240125205218-1f4bbc51befe cloud.google.com/go@v0.112.0
cloud.google.com/go/accesscontextmanager@v1.8.5 cloud.google.com/go@v0.111.0
cloud.google.com/go/aiplatform@v1.60.0 cloud.google.com/go@v0.112.0
cloud.google.com/go/apigateway@v1.6.5 cloud.google.com/go@v0.111.0
cloud.google.com/go/apigeeregistry@v0.8.3 cloud.google.com/go@v0.111.0
cloud.google.com/go/appengine@v1.8.5 cloud.google.com/go@v0.111.0
cloud.google.com/go/artifactregistry@v1.14.7 cloud.google.com/go@v0.111.0
[...]

PR #2614, opened 2024-05-31
fix: add another temporary dep on genproto

codyoss:
I believe the soure for these issue is opencensus outdated imports of really old version of our libraries, genproto. Once we drop that dep we should test removing these workarounds.

Updates: #2559
Fixes: #2613


quartzmo (approved):
Thanks!

Issue #2601, opened 2024-05-23
Properly support "strong typed maps" in generator


codyoss:
Currently the generator won't generate a type for the following code when the type is a request or response to a method: 

```
"GoogleProtobufStruct": {
      "id": "GoogleProtobufStruct",
      "description": "description",
      "type": "object",
      "additionalProperties": {
        "type": "any",
        "description": "Properties of the object."
      }
    },
```

Code fails on something like: https://github.com/googleapis/google-api-go-client/blob/64f40ff95a3fd126680805e762b0824f69f0e218/google-api-go-generator/gen.go#L1362-L1363

All other language generators support this.

Internal Bug: 336571031

PR #2632, opened 2024-06-12
feat: support structpb.Struct as req/resp

codyoss:
There are some APIs that have started to use this type for the request and/or respsonse. Similar to how we had to specially handle protos HTTP body we need a good translation for this type as well. `map[string]any` seemed like the best fit as that is the input needed to create a `Struct`. The other choice would have been a `googleapis.RawMessage`. RawMessage is used today when a field would be of type Struct, but this is a less convient type, and less precise type, to use than a map directly.

Fixes: #2601


quartzmo (approved):
LGTM

PR #2643, opened 2024-06-19
fix(option): skip nil bytes for CredentialsJSON

lexcao:
Closes https://github.com/googleapis/google-api-go-client/issues/2647
Nil bytes was treated to empty bytes, which causes failure on empty JSON. 
I think we can improve this by ignoring nil bytes.

### DIFF
--- a/option/option.go
+++ b/option/option.go
@@ -65,6 +65,9 @@ func WithCredentialsJSON(p []byte) ClientOption {
 type withCredentialsJSON []byte
 
 func (w withCredentialsJSON) Apply(o *internal.DialSettings) {
+	if w == nil {
+		return
+	}
 	o.CredentialsJSON = make([]byte, len(w))
 	copy(o.CredentialsJSON, w)
 }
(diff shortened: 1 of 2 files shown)


quartzmo:
@lexcao Please create a [bug report](https://github.com/googleapis/google-api-go-client/issues/new/choose) (or possibly, feature request, depending on how you view this), we need that before we can move forward with this PR.

quartzmo:
In your issue, please explain why you think adding the conditional to avoid setting the `nil` value needs to be in this package, and not just in the user code such as in your PR https://github.com/authzed/spicedb/pull/1948.

lexcao:
Hi @quartzmo 
Thanks for reviewing, let me create an issue to provide more context first!

lexcao:
Hi @quartzmo 
I have created an issue https://github.com/googleapis/google-api-go-client/issues/2647.
Please help review.
Thanks!

codyoss:
I am going to close this as I opened up an alternative PR, thank you for the report though!

PR #2657, opened 2024-06-27
fix(gensupport): wrap chunk upload err for retries

tritone:
This makes it easier to debug whether and/or how many retries happened before a failure with a chunk upload.

### DIFF
--- a/internal/gensupport/resumable.go
+++ b/internal/gensupport/resumable.go
@@ -171,6 +171,10 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
 			if resp != nil && resp.Body != nil {
 				resp.Body.Close()
 			}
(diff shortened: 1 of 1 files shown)


tritone:
@tulsishah I can't add you as a reviewer in this repo but PTAL. This should help us understand how many retries are able to happen before the deadline.

codyoss (approved):

Tulsishah (approved):
Should we also add this error message [here](https://github.com/Tulsishah/google-cloud-go/blob/c155efe42843b1fedb6fb20aa5af96dad44a9f47/storage/invoke.go#L73)?

tritone:
> Should we also add this error message [here](https://github.com/Tulsishah/google-cloud-go/blob/c155efe42843b1fedb6fb20aa5af96dad44a9f47/storage/invoke.go#L73)?

This was already added in googleapis/google-cloud-go#9767

PR #2670, opened 2024-07-03
fix: allow ForceSendFields to work for map types

codyoss:
The way we generate map types today is map[string]T, where T is a non-pointer type. This means the pointer receiver version of MarshalJSON does not fulfill the Marshaler interface. This is turn means that MarshalJSON is not called in these cases so ForceSendFields overrides do not work. To fix this we can switch the impl to a value receiver which fulfills both value and pointer types. This is not a breaking change for the API, but there is a slight behaviour change if someone was calling MarshalJSON directly. Nil *T's that call MarshalJSON directly will now panic instead of returning JSON `null`. This seems like unlikely usage and worth breaking to fix a bug in common usage.

Internal Bug: 349580049


noahdietz (approved):

tritone (approved):

diviner524:
> This means the value receiver version of MarshalJSON does not fulfill the Marshaler interface.

nit: The description looks to be a bit misleading. I think you mean the pointer receiver version of MarshalJSON.

diviner524:
/lgtm

Thanks for the quick fix!

shollyman (approved):

PR #2690, opened 2024-07-15
fix(transport/grpc): retain UserAgent option with new auth stack

noahdietz:
The only code that converted the `option.WithUserAgent` to a gRPC DialOption `WithUserAgent` was in the `dial` function, which was bypassed when leveraging the new auth library that handles most of the important dialing set up re:credentials.

To fix this, I added a new helper specifically for use in `dialPoolNewAuth` that prepares dial options based on the given `DialSettings` - `prepareDialOpionsNewAuth`. I could've added the same logic to anywhere in `dialPoolNewAuth` to fix it, but I felt I needed to signal that this is the method to be modifying gRPC DialOptions in so that in the future, we can centralize further gRPC DialOption manipulation rather than just "wherever" like in the "old" `dial`.

Also stored the `grpctransport.Dial` function into a global variable so that unit tests could spoof it to check on the gRPC DialOptions that were provided - just like the "old" `dial` did for testing.

Updates https://github.com/googleapis/google-cloud-go/issues/10550 but needs to be released here, integrated there, and released there.

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -53,6 +53,9 @@ var logRateLimiter = rate.Sometimes{Interval: 1 * time.Second}
 // Assign to var for unit test replacement
 var dialContext = grpc.DialContext
 
+// Assign to var for unit test replacement
+var dialContextNewAuth = grpctransport.Dial
+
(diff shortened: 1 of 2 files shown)


codyoss (approved):
Thanks!

PR #2696, opened 2024-07-24
fix(idtoken): include email in Payload

rolandshoemaker:
The Payload type was missing the email field, which is useful when validating IAP JWTs, in order to dissuade using the x-goog-authenticated-user-email header. Unclear to me why it was originally omitted.


codyoss:
I think for now at least we are considering this WAI with what other langs are doing. Pointed this out offline, but this info is accessible via `Payload.Claims`

PR #2717, opened 2024-08-01
fix(transport): disable automatic universe domain check

quartzmo:


codyoss (approved):

codyoss:
Looks good besides the test, thanks

PR #2745, opened 2024-08-22
fix(transport): Add EarlyTokenRefresh in auto detect options

bhshkh:
**Issue**:
While running the Bigtable client on GCE instance, token is not expired early. 

**Cause**:
computeTokenProvider sets ExpireEarly to EarlyTokenRefresh but there is no way to pass this option in the gtransport.DialPool method used by bigtable.go

https://github.com/googleapis/google-cloud-go/blob/1afb9ee1ee9de9810722800018133304a0ca34d1/auth/credentials/compute.go#L40-L45

**Fix**:
Set EarlyTokenRefresh while auto detecting credentials

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -38,8 +38,14 @@ import (
 	_ "google.golang.org/grpc/balancer/grpclb"
 )
 
-// Check env to disable DirectPath traffic.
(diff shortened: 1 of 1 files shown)


bhshkh:
This is being fixed in the new auth library itself. https://github.com/googleapis/google-cloud-go/pull/10732
This PR is no longer needed

PR #2758, opened 2024-08-27
feat(storage): allow Accept-Encoding header

### LINKED ISSUE #2757: gensupport/storage: allow Accept-Encoding header
**Is your feature request related to a problem? Please describe.**
In Storage, we use the `Accept-Encoding` header to allow users to download the zipped version of their data, see [`ReadCompressed`](https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.ReadCompressed). We cannot support this in JSON because the header is disallowed. 

**Describe the solution you'd like**
Allow the header, at least for storage media downloads. This requires changes to the generated code.

**Describe alternatives you've considered**
Allow the header in general. This can be done in gensupport.

BrennaEpp:
Fixes #2757 

Code needs to be regenerated to match up with the additional  parameters in gensupport.


quartzmo:
kokoro error:

```sh
apigee/v1/apigee-gen.go:12017:52: not enough arguments in call to gensupport.SendRequest
	have (context.Context, *"net/http".Client, *"net/http".Request)
	want (context.Context, *"net/http".Client, *"net/http".Request, bool)
```

quartzmo:
In order to get a passing Kokoro presubmit check for this PR, you may want to regenerate all of the clients locally and then commit them in your branch, as @codyoss and I usually do. (Recent example in #2670.) You can do it by running `make cached` in the `google-api-go-generator` directory. You may also need to run tests locally with the `update_golden` flag if the golden files are in scope for this change. (See `gen_test.go` for how that works.)

BrennaEpp:
> In order to get a passing Kokoro presubmit check for this PR, you may want to regenerate all of the clients locally and then commit them in your branch, as @codyoss and I usually do. (Recent example in #2670.) You can do it by running `make cached` in the `google-api-go-generator` directory. You may also need to run tests locally with the `update_golden` flag if the golden files are in scope for this change. (See `gen_test.go` for how that works.)

I updated everything, but am still seeing errors locally (I can't see the kokoro build):
```
--- FAIL: TestNewClient (0.00s)
    dial_test.go:26: got *oauth2.Transport, want: *httptransport.authTransport
FAIL
FAIL    google.golang.org/api/transport/http    0.347s
```
This seems unrelated to the changes I made, can you help me fix?

codyoss:
@BrennaEpp is this a change we are still wanting to land?

shollyman:
stale PR

PR #2767, opened 2024-09-02
Use QuotaProject in detect options

michalskalski:
To help lookup quota project id with proper sources priority order.

It depends on https://github.com/googleapis/google-cloud-go/pull/10805


codyoss:
This is taken care of, closing.

PR #2773, opened 2024-09-05
fix(transport): set UniverseDomain in http.NewClient for new auth

quartzmo:
* Set value when GOOGLE_API_GO_EXPERIMENTAL_ENABLE_NEW_AUTH_LIB=true

### DIFF
--- a/transport/http/dial.go
+++ b/transport/http/dial.go
@@ -130,6 +130,7 @@ func newClientNewAuth(ctx context.Context, base http.RoundTripper, ds *internal.
 			DefaultScopes:           ds.DefaultScopes,
 			SkipValidation:          skipValidation,
 		},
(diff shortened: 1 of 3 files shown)


transport/http/dial.go
```diff
@@ -130,6 +130,7 @@ func newClientNewAuth(ctx context.Context, base http.RoundTripper, ds *internal.
 			DefaultScopes:           ds.DefaultScopes,
 			SkipValidation:          skipValidation,
 		},
+		UniverseDomain: ds.UniverseDomain,
```
codyoss, line 133: Do we need something similar for gRPC?

quartzmo: Fixed, thanks!

codyoss (approved):

PR #2798, opened 2024-09-24
fix(transport): restore support for GOOGLE_CLOUD_UNIVERSE_DOMAIN env var

quartzmo:
The `GOOGLE_CLOUD_UNIVERSE_DOMAIN `env variable no longer works.

If I use `google.golang.org/api@v0.191.0` I can start my example with `GOOGLE_CLOUD_UNIVERSE_DOMAIN=example.com` and it works without specifying `WithUniverseDomain` option explicitly.

When I use `google.golang.org/api@v0.197.0` and `cloud.google.com/go/auth@v0.9.4` I get error when I am using this env, if I don't specify `WithUniverseDomain` explicitly.

Error: the configured universe domain ("googleapis.com") does not match the universe domain found in the credentials ("example.com"). If you haven't configured the universe domain explicitly, "googleapis.com" is the default

### DIFF
--- a/transport/http/dial.go
+++ b/transport/http/dial.go
@@ -30,6 +30,9 @@ import (
 	"google.golang.org/api/transport/http/internal/propagation"
 )
 
+// package var for unit test substitution
+var newClient = httptransport.NewClient
+
 // NewClient returns an HTTP client for use communicating with a Google cloud
 // service, configured with the given ClientOptions. It also returns the endpoint
 // for the service as specified in the options.
@@ -107,7 +110,7 @@ func newClientNewAuth(ctx context.Context, base http.RoundTripper, ds *internal.
 	if ds.RequestReason != "" {
 		headers.Set("X-goog-request-reason", ds.RequestReason)
 	}
-	client, err := httptransport.NewClient(&httptransport.Options{
+	client, err := newClient(&httptransport.Options{
 		DisableTelemetry:      ds.TelemetryDisabled,
 		DisableAuthentication: ds.NoAuth,
 		Headers:               headers,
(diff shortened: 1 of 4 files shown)


transport/http/dial_test.go
```diff
@@ -26,3 +31,21 @@ func TestNewClient(t *testing.T) {
 		t.Fatalf("got %s, want: %s", got, want)
 	}
 }
+
+func TestNewClient_NewAuthUniverseDomain(t *testing.T) {
+	os.Setenv("GOOGLE_CLOUD_UNIVERSE_DOMAIN", "example.com")
```
codyoss, line 36: nit: use https://pkg.go.dev/testing#T.Setenv. then you don't need to do this cleanup game.

transport/http/dial.go
```diff
@@ -130,7 +133,7 @@ func newClientNewAuth(ctx context.Context, base http.RoundTripper, ds *internal.
 			DefaultScopes:           ds.DefaultScopes,
 			SkipValidation:          skipValidation,
 		},
-		UniverseDomain: ds.UniverseDomain,
+		UniverseDomain: ds.GetUniverseDomain(),
```
codyoss, line 136: Do we need to do this in the new library as well for those that use it directly?

quartzmo: Yes, `DialSettings.UniverseDomain` usages should be replaced anywhere they occur with `DialSettings.GetUniverseDomain()` in order to support the env var.

transport/http/dial_test.go
quartzmo: Fixed, thanks!

transport/http/dial.go
codyoss: The new lib does not have dialsettings, though. But it does allow someone to explicitly set this value. But if they don't, should we honor this there? If we did honor this in the new library and had the current code here I think the bug would be fixed for both, right?

quartzmo: I will update this in `grpc/dial.go` in this PR, thank you.

quartzmo:
Closing in favor of a fix in the new `auth` module:

https://github.com/googleapis/google-cloud-go/pull/10915

PR #2810, opened 2024-10-02
feat(storagetransfer): Support FederatedIdentifyConfig for Azure Blob Storage

venky999:
Storage Transfer Service: Support FederatedIdentifyConfig for Azure Blob Storage

### DIFF
--- a/storagetransfer/v1/storagetransfer-api.json
+++ b/storagetransfer/v1/storagetransfer-api.json
@@ -754,7 +754,7 @@
       "properties": {
         "azureCredentials": {
           "$ref": "AzureCredentials",
-          "description": "Required. Input only. Credentials used to authenticate API requests to Azure. For information on our data retention policy for user credentials, see [User credentials](/storage-transfer/docs/data-retention#user-credentials)."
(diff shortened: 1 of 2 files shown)


codyoss:
These files are pulled from the public discovery doc listing -- we do not allow hand editing them. Please make a request to the API service if you wish to see a feature added to a service.

venky999:
> These files are pulled from the public discovery doc listing -- we do not allow hand editing them. Please make a request to the API service if you wish to see a feature added to a service.

@codyoss May I know where the go api is .. because the api reference here https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl#L18

codyoss:
This file is a cached version of what is being served at: https://storagetransfer.googleapis.com/$discovery/rest?version=v1

If there you think there is something missing from this definition it is best to reach out to that teams support page: https://cloud.google.com/storage-transfer/docs/support. This repo just generates clients from the public spec.

venky999:
Thank you @codyoss I see the API is missing this feature but I am not sure how to raise the support

PR #2818, opened 2024-10-08
feat: regenerated firebase remote config client with parameter groups support

hrvadl:
Added missing `parameterGroups` field to the remote config client


codyoss:
These files are pulled from the public discovery doc listing -- we do not allow hand editing them. Please make a request to the API service if you wish to see a feature added to a service.

PR #2854, opened 2024-11-01
feat(option): add support for user loggers

codyoss:
Also add an intenral option that will be used by our clients to extract a logger from the options, or a default logger if one is not specified.


codyoss:
Requires https://github.com/googleapis/gax-go/pull/380

PR #2865, opened 2024-11-08
feat(gensupport): per-chunk transfer timeout configs

Tulsishah:
feat(gensupport): per-chunk transfer timeout configs

Allow users to configure the per-chunk transfer timeout for retries
that's used during resumable uploads.

Needs to be exposed via the manual layer for storage.

Tested the feature(with default timeout and with some random value) with storagetestbench server emulator.


internal/gensupport/retry.go
```diff
@@ -26,6 +26,8 @@ type Backoff interface {
 var (
 	// Default per-chunk deadline for resumable uploads.
 	defaultRetryDeadline = 32 * time.Second
+	// Default per-chunk transfer timeout for resumable uploads.
+	defaultTransferTimeout = 8 * time.Second
```
tritone, line 30: Default should be no timeout.

internal/gensupport/resumable_test.go
```diff
@@ -280,7 +280,7 @@ func TestCancelUploadBasic(t *testing.T) {
 	defer func() { backoff = oldBackoff }()
 
 	res, err := rx.Upload(ctx)
-	if err != context.Canceled {
+	if strings.Contains(err.Error(), "context cancelled") {
```
tritone, line 283: Why is this change needed? Can you use errors.Is?

internal/gensupport/resumable.go
```diff
@@ -241,15 +253,23 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
...
+			defer cancel()
+
+			resp, err = rx.transferChunk(rCtx)
 
 			var status int
 			if resp != nil {
 				status = resp.StatusCode
 			}
 
+			if errors.Is(err, context.DeadlineExceeded) {
```
tritone, line 266: How do we distinguish between a timeout in ctx vs rCtx?

I think we'd only want to retry if we get deadline exceeded from rCtx, but ctx is still good.

internal/gensupport/resumable_test.go
Tulsishah: Done.

internal/gensupport/resumable.go
Tulsishah: I feel below condition is enough to check this `I think we'd only want to retry if we get deadline exceeded from rCtx, but ctx is still good.` Removed that condition.

internal/gensupport/retry.go
Tulsishah: Done

googleapi/googleapi.go
```diff
@@ -259,6 +259,20 @@ func ChunkSize(size int) MediaOption {
...
 }
 
+type chunkTransferTimeoutOption time.Duration
+
+func (cd chunkTransferTimeoutOption) setOptions(o *MediaOptions) {
+	o.ChunkTransferTimeout = time.Duration(cd)
+}
+
+// ChunkTransferTimeout returns a MediaOption which sets a per-chunk
+// transfer timeout. If a single chunk has been attempting to upload for longer than
```
tritone, line 269: Clarify: `// transfer timeout for resumable uploads.`

googleapi/googleapi.go
```diff
@@ -259,6 +259,20 @@ func ChunkSize(size int) MediaOption {
...
+// this time then the old req got canceled and retried.
+// The default value is 0s. This ensures that existing clients won't experience any changes in behavior.
```
tritone, line 271: I would drop the sentence about existing clients; instead something like:

`// The default is no timeout for the request.`

internal/gensupport/resumable.go
```diff
@@ -188,6 +192,14 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
 	// Configure retryable error criteria.
 	errorFunc := rx.Retry.errorFunc()
 
+	// Configure per-chunk transfer timeout.
+	var transferTimeout time.Duration
+	if rx.ChunkTransferTimeout != 0 {
+		transferTimeout = rx.ChunkTransferTimeout
+	} else {
+		transferTimeout = defaultTransferTimeout
```
tritone, line 200: I think since the default is zero, we don't need this if/else?

internal/gensupport/retry.go
```diff
@@ -26,6 +26,8 @@ type Backoff interface {
 var (
 	// Default per-chunk deadline for resumable uploads.
 	defaultRetryDeadline = 32 * time.Second
+	// Default per-chunk transfer timeout for resumable uploads.
+	defaultTransferTimeout = 0 * time.Second
```
tritone, line 30: I think we don't need a variable for this given that zero is the language default for time.Duration.

internal/gensupport/resumable.go
```diff
@@ -241,13 +253,31 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
...
 			}
 
-			resp, err = rx.transferChunk(ctx)
+			var rCtx context.Context
+			var cancel context.CancelFunc
+
+			rCtx = ctx
+			if transferTimeout != 0 {
+				rCtx, cancel = context.WithTimeout(ctx, transferTimeout)
+				defer cancel()
```
tritone, line 262: I'm not sure this is a big deal practically speaking, but I'm not sure `defer cancel()` is what you want here, since that doesn't return until the function returns. So we might have a bunch of contexts stack up over the life of this loop.

Maybe we can explicitly cancel after transferChunk returns instead?

tritone (commented):
Few more points here.

googleapi/googleapi.go
Tulsishah: Done

internal/gensupport/resumable.go
Tulsishah: I kept it this way to make it easier to update the default value from 0 to something else in the future.  Anyways updated the changes as you said.

internal/gensupport/retry.go
Tulsishah: I kept it this way to make it easier to update the default value from 0 to something else in the future.  Anyways updated the changes as you said.

internal/gensupport/resumable.go
Tulsishah: Good catch. added the cancel when context is done.

googleapi/googleapi.go
```diff
@@ -259,6 +259,20 @@ func ChunkSize(size int) MediaOption {
...
+type chunkTransferTimeoutOption time.Duration
+
+func (cd chunkTransferTimeoutOption) setOptions(o *MediaOptions) {
+	o.ChunkTransferTimeout = time.Duration(cd)
+}
+
+// ChunkTransferTimeout returns a MediaOption which sets a per-chunk
+// transfer timeout for resumable uploads. If a single chunk has been
+// attempting to upload for longer than this time then the old req got canceled and retried.
+// The default is no timeout for the request. This ensures that existing clients won't experience any changes in behavior.
```
tritone, line 271: You can remove the sentence about existing clients.

internal/gensupport/resumable.go
```diff
@@ -241,13 +245,32 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
...
 			if resp != nil {
 				status = resp.StatusCode
 			}
 
+			// The upload should be retried if the rCtx is canceled due to a timeout.
+			select {
+			case <-rCtx.Done():
+				if rCtx.Err() == context.DeadlineExceeded {
+					// Cancel the context for rCtx
+					cancel()
```
tritone, line 268: I think this would panic with chunkTransferTimeout = 0 because the function was never set to anything. Did you test this case?

internal/gensupport/resumable.go
```diff
@@ -241,13 +245,32 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
 			default:
 			}
 
-			resp, err = rx.transferChunk(ctx)
+			var rCtx context.Context
```
tritone, line 248: Can you add a comment above saying what rCtx is for?

internal/gensupport/resumable.go
```diff
@@ -241,13 +245,32 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
...
 
 			var status int
 			if resp != nil {
 				status = resp.StatusCode
 			}
 
+			// The upload should be retried if the rCtx is canceled due to a timeout.
+			select {
+			case <-rCtx.Done():
+				if rCtx.Err() == context.DeadlineExceeded {
```
tritone, line 266: Use errors.Is()

tritone (commented):
Few more comments here.

googleapi/googleapi.go
Tulsishah: Done

internal/gensupport/resumable.go
Tulsishah: Yes I have done manual testing, it is not having panic. I have assign rCtx=ctx and override the rCtx value in case of transfertimeout non zero. If it is zero then rCtx=ctx.

codyoss (commented):
LGTM to me overall, I will wait for @tritone to fully sign-off. I will say this function is starting to get a bit long and complex. I would consider breaking up the loop body into some different functions can be better documented. Then you can use things like defer again too.

tritone (approved):

tritone:
> LGTM to me overall, I will wait for @tritone to fully sign-off. I will say this function is starting to get a bit long and complex. I would consider breaking up the loop body into some different functions can be better documented. Then you can use things like defer again too.

It's a good point; I can do this in a subsequent PR.

codyoss (approved):

PR #2884, opened 2024-11-21
fix(transport): copy map in grpcTokenSource.GetRequestMetadata

quartzmo:
fixes: #11161

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -379,11 +379,15 @@ type grpcTokenSource struct {
 }
 
 // GetRequestMetadata gets the request metadata as a map from a grpcTokenSource.
-func (ts grpcTokenSource) GetRequestMetadata(ctx context.Context, uri ...string) (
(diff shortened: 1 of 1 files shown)


quartzmo:
This PR is not needed: The grpc function that is called returns a new map:

```
// GetRequestMetadata gets the request metadata as a map from a TokenSource.
func (ts TokenSource) GetRequestMetadata(ctx context.Context, _ ...string) (map[string]string, error) {
    token, err := ts.Token()
    if err != nil {
        return nil, err
    }
    ri, _ := credentials.RequestInfoFromContext(ctx)
    if err = credentials.CheckSecurityLevel(ri.AuthInfo, credentials.PrivacyAndIntegrity); err != nil {
        return nil, fmt.Errorf("unable to transfer TokenSource PerRPCCredentials: %v", err)
    }
    return map[string]string{
        "authorization": token.Type() + " " + token.AccessToken,
    }, nil
}
```

PR #2891, opened 2024-11-26
feat(option/internaloption): add AuthCreds helper

quartzmo:
refs: googleapis/google-cloud-go#11191

### DIFF
--- a/option/internaloption/internaloption.go
+++ b/option/internaloption/internaloption.go
@@ -6,8 +6,10 @@
 package internaloption
 
 import (
+	"context"
 	"log/slog"
 
+	"cloud.google.com/go/auth"
 	"github.com/googleapis/gax-go/v2/internallog"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/internal"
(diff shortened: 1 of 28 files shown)


transport/dial.go
```diff
@@ -46,3 +47,30 @@ func Creds(ctx context.Context, opts ...option.ClientOption) (*google.Credential
...
+// following options to [cloud.google.com/go/auth/credentials.DetectDefault] and
+// returns the result:
+//
+// * [option.WithAudiences]
+// * [option.WithCredentialsFile]
+// * [option.WithCredentialsJSON]
+// * [option.WithScopes]
+// * [option/internaloption.WithDefaultScopes] (internal use only)
+// * [option/internaloption.EnableJwtWithScope] (internal use only)
+func AuthCreds(ctx context.Context, opts ...option.ClientOption) (*auth.Credentials, error) {
```
codyoss, line 70: I understand the reasoning for this PR and I think we do need these sort of APIs, but we will need at least two more of this flavor as well -- one for calling new http and one for new grpc libs(from generated code). I don't think this should be a user facing API, although it needs to be public and needs to be in this module. Could we put this in an internaloption type of package instead? Thoughts?

quartzmo: In my opinion, the need for a helper like this is a symptom of poor design elsewhere. (1. The storage client never should have used `option.ClientOption` for its own options, since it can't inspect the contents. 2. Why does `option.ClientOption` not expose its contents!?!) Given the present situation, I like the idea of this helper being located in an "internaloption" package with an EXPERIMENTAL label. Hopefully then we can solve the deeper design problem and remove this helper.

This helper is immediately needed in order to migrate the Storage to client to the new auth lib (see https://github.com/googleapis/google-cloud-go/pull/11191). Maybe the need for the other helpers (one for calling new http and one for new grpc libs) could be solved in a different way?

codyoss: > This helper is immediately needed in order to migrate the Storage to client to the new auth lib (see https://github.com/googleapis/google-cloud-go/pull/11191). Maybe the need for the other helpers (one for calling new http and one for new grpc libs) could be solved in a different way?

I think it will have to be the same thing if we want to use client options to call directly into new APIs. We need to do that translation somewhere. And because, like you mentioned, the options not sharing their internal state it has to be somewhere in this module today.

> , I like the idea of this helper being located in an "internaloption" package with an EXPERIMENTAL label.

It is good as spot as any I suppose. We kinda had to do a hack like this for the logging work for similar reasons

option/internaloption/internaloption.go
```diff
@@ -228,3 +230,33 @@ func GetLogger(opts []option.ClientOption) *slog.Logger {
...
+// * [option.WithAudiences]
+// * [option.WithCredentialsFile]
+// * [option.WithCredentialsJSON]
+// * [option.WithScopes]
+// * [option/internaloption.WithDefaultScopes] (internal use only)
+// * [option/internaloption.EnableJwtWithScope] (internal use only)
+//
+// This function should only be used internally by generated clients. This is an
+// EXPERIMENTAL API and may be changed or removed in the future.
+func AuthCreds(ctx context.Context, opts ...option.ClientOption) (*auth.Credentials, error) {
```
codyoss, line 256: nit: instead of ...opts take a slice. Callers will always have a slice at call time.

codyoss (approved):

Issue #2543, opened 2024-04-24
Import broken when using only androidpublisher/v3


CoreFloDev:
Hi,

I am using only google.golang.org/api/androidpublisher/v3 on my application to manage play store subscription. Since the v0.174.0 when the dependency to app engine was removed, I got import errors:
```
api          | 2024/04/24 13:32:12 Error while building:
api          |  go: downloading google.golang.org/api v0.176.1
api          | go: downloading golang.org/x/sync v0.7.0
api          | go: downloading cloud.google.com/go/auth v0.3.0
api          | go: downloading cloud.google.com/go/auth/oauth2adapt v0.2.2
api          | go: downloading cloud.google.com/go/compute v1.24.0
api          | go: downloading cloud.google.com/go v0.26.0
api          | /go/pkg/mod/cloud.google.com/go/auth@v0.3.0/internal/internal.go:31:2: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
api          |  cloud.google.com/go v0.26.0 (/go/pkg/mod/cloud.google.com/go@v0.26.0/compute/metadata)
api          |  cloud.google.com/go/compute/metadata v0.3.0 (/go/pkg/mod/cloud.google.com/go/compute/metadata@v0.3.0)
```

I don't get how to address that issue, the only workaround I have found to far is too keep cloud.google.com/go/compute v1.25.1 // indirect

Unfortunately it gets removed at every go work sync or go mod tidy as I don't use it on my module.

Is there a better solution?

Have a nice day,
Florent

codyoss:
I just made a new project with this code which builds without issue:

go
```
package main

import (
	"context"

	"google.golang.org/api/androidpublisher/v3"
)

func main() {
	ctx := context.Background()
	androidpublisher.NewService(ctx)
}
```

If you are using up-to-date dependencies here it should not be an issue. Please share your go.mod and/or upgrade `cloud.google.com/go` to latest. The version mentioned in your comment is from 2018.

CoreFloDev:
I don't get why it mention `cloud.google.com/go`

it is not required on the go.mod I have
```go.mod
go 1.22.2

require (
	github.com/99designs/gqlgen v0.17.45
	github.com/go-chi/chi/v5 v5.0.12
	github.com/go-webauthn/webauthn v0.10.2
	github.com/google/wire v0.6.0
	github.com/jackc/pgx/v5 v5.5.5
[...]

PR #2908, opened 2024-12-10
fix: remove dep hacks for old cycle deps

codyoss:
I believe our dep cycle issues should have been fixed in #2907.
Verified issue in #2543 is fixed with this change.

Fixes: #2559


quartzmo (approved):
Nice cleanup!

Issue #2917, opened 2024-12-13
Google Safe Browsing v5 missing v5alpha1 new features


JILeXanDR:
#### Steps to reproduce

In docs https://developers.google.com/safe-browsing/reference/rest there are new methods with prefix **v5alpha1** but they are missing in the generated code https://github.com/googleapis/google-api-go-client/blob/main/safebrowsing/v5/safebrowsing-api.json

Same here https://safebrowsing.googleapis.com/$discovery/rest?version=v5 when checking via discovery tool.

codyoss:
These libraries are generated by reading the public discovery for the service. If the features are not there we can't generate. I suggest reaching out via the services support page to report this. Hope that helps.

JILeXanDR:
@codyoss Hi, I mentioned wrong discovery link, it's `v5`, but new API is listed under `v5alpha1`.

https://safebrowsing.googleapis.com/$discovery/rest?version=v5alpha1

This is already available and contains all needed methods to include in this repo.

PR #2932, opened 2024-12-19
fix: add missing api safebrowsing:v5alpha1

JILeXanDR:
Has started as this issue https://github.com/googleapis/google-api-go-client/issues/2917.

This MR adds missing API for safebrowsing API v5alpha1 version. But not sure should it be `preferred: true` or not since it's alpha.

### DIFF
--- a/api-list.json
+++ b/api-list.json
@@ -5544,6 +5544,21 @@
     },
     {
       "kind": "discovery#directoryItem",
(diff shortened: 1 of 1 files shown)


codyoss:
Sorry we don't accept contributions to this file. This file is generated by looking at the list of public apis here: https://www.googleapis.com/discovery/v1/apis

PR #2964, opened 2025-01-16
feat: add error helper that does not read the response body

bjohnso5:
Adds a new `CheckResponseWithBody` variant to allow returning `*googleapi.Error` details for responses where the body may have already been read.

Fixes https://github.com/googleapis/google-cloud-go/issues/11458 (Part 1 of 2).


googleapi/googleapi.go
```diff
@@ -145,22 +145,40 @@ func CheckResponse(res *http.Response) error {
 	}
 	slurp, err := io.ReadAll(res.Body)
```
quartzmo, line 146: @codyoss Did you ever notice that if `err` from `ReadAll` is non-nil, it is never used (lines 150-154, below)?

codyoss: I believe the intention of this function is to provide more context to the error in all cases, that is why it is never returned directly. We can at least always append some HTTP info.

codyoss (approved):

codyoss:
Thanks for the detailed report and contribution!

bjohnso5:
@codyoss is there anything I can do re: the kokoro failures? It's not clear what the failure actually was based on the output I see.

PR #2966, opened 2025-01-17
fix(internal/gensupport): close resp body only on discarding resp (resumableupload)

BrennaEpp:
internal, see: b/389767299


internal/gensupport/resumable.go
```diff
@@ -256,6 +256,10 @@ func (rx *ResumableUpload) Upload(ctx context.Context) (resp *http.Response, err
 				rCtx, cancel = context.WithTimeout(ctx, rx.ChunkTransferTimeout)
 			}
 
+			if resp != nil && resp.Body != nil {
```
tritone, line 259: Add a comment about why this is necessary and why it happens at this point in the loop.

codyoss (approved):

tritone (approved):

PR #2975, opened 2025-01-22
feat(option/internaloption): add new allowHardBoundTokens option

yamandabbagh:
Add allowHardBoundTokens option to the internaloption. This option will be used internally only to allow auto-generated clients to request a hard-bound tokens. Hard-bound tokens are tokens that include binding that must be enforced regardless of the IAM policy.


quartzmo:
Do not merge was applied pending review of the design doc, which @codyoss approved.

option/internaloption/internaloption.go
```diff
@@ -186,6 +186,21 @@ func (w enableJwtWithScope) Apply(o *internal.DialSettings) {
 	o.EnableJwtWithScope = bool(w)
 }
 
+// AllowHardBoundTokens returns a ClientOption that allows libraries to request a hard-bound token.
```
quartzmo, line 189: "allow" implies agency, or the ability for the client or auth library to decide whether or not to "request a hard-bound token". Can you briefly describe here under which circumstances the decision to request a hard-bound token will be made?

option/internaloption/internaloption.go
```diff
@@ -186,6 +186,21 @@ func (w enableJwtWithScope) Apply(o *internal.DialSettings) {
...
+//
```
quartzmo, line 190: Please enumerate the possible values and their definitions in this doc comment, thank you.

quartzmo (changes requested):

option/internaloption/internaloption.go
yamandabbagh: Updated the comment to include more details. The decision to request hard-bound tokens will depend on the transport used. If, for example, AllowHardBoundTokens = {MTLS_S2A}, and the connection between the client and the Google APIs is actually using mTLS and S2A then the token requested will be a hard-bound token.

option/internaloption/internaloption.go
yamandabbagh: Added! it is much clearer now, thank you for the suggestion!

option/internaloption/internaloption.go
```diff
@@ -186,6 +186,32 @@ func (w enableJwtWithScope) Apply(o *internal.DialSettings) {
 	o.EnableJwtWithScope = bool(w)
 }
 
+// AllowHardBoundTokens returns a ClientOption that allows libraries to request a hard-bound token.
+// Obtaining hard-bound tokens requires the connection to be established using either
+// DirectPath or mutual TLS (mTLS) with S2A.
```
quartzmo, line 191: What is the relationship between DirectPath and the two options listed below? Is there a connection between DirectPath and ALTS? What is ALTS? Are there any docs you can link? ("For more information on ALTS, see ...").

I understand that some features may not be widely publicized, but these are public docs. If we mention a feature, some users will be curious and want to know enough about the topic just to be sure it does not interest them.

yamandabbagh: Actually, the DirectPath is the Google specific name for the cloud to prod connection that uses ALTS. So I think dropping the DirectPath and keeping the ALTS with a link to more info about ALTS is sufficient. WDYT?

codyoss (approved):

yamandabbagh:
kokoro rerun

yamandabbagh:
kokoro rerun

codyoss:
The failure is do to the file needing to be formatted:

```
+ gofmt -s -d -l .
+ tee /dev/stderr
+ read
option/internaloption/internaloption.go
```

PR #2978, opened 2025-01-23
docs(option): add warning about externally-provided credentials

quartzmo:
cc @sai-sunder-s


option/option.go
```diff
@@ -44,13 +44,29 @@ func (w withCredFile) Apply(o *internal.DialSettings) {
...
+// Important: If you accept a credential configuration (credential
+// JSON/File/Stream) from an external source for authentication to Google
+// Cloud Platform, you must validate it before providing it to any Google
+// API or library. Providing an unvalidated credential configuration to
+// Google APIs can compromise the security of your systems and data. For
+// more information, refer to [Validate credential configurations from
+// external sources](https://cloud.google.com/docs/authentication/external/externally-sourced-credentials).
+//
 // Deprecated: Use WithCredentialsFile instead.
 func WithServiceAccountFile(filename string) ClientOption {
```
sai-sunder-s, line 71: does this method accept only service account or any file?

quartzmo: It is a deprecated method. It was probably deprecated because the name was too limited. It now delegates to `WithCredentialsFile`.

codyoss (approved):

option/option.go
sai-sunder-s: Got it

PR #2997, opened 2025-02-06
chore: remove unnecessary err != nil check

anasmuhmd:


codyoss:
Your right this does look like dead code but this file is generated and can't be updated directly. Would you mind filing an issue related to this change.

anasmuhmd:
@codyoss https://github.com/googleapis/google-api-go-client/issues/3001

Issue #3009, opened 2025-02-11
go.mod patch version is forcing toolchain, breaking builds


matthewhartstonge:
Hi Team,

I noticed that RenovateBot added a patch released that pushed minimum requirements of Go up to 1.22.11. This patch version was introduced via revonatebot https://github.com/googleapis/google-api-go-client/pull/2992/files due to https://github.com/googleapis/go-genproto/pull/1185

This has ended up proliferating through the API SDKs and breaking CI/dependency upgrade builds.

Can the version requirement be relaxed to `1.22` to remain compatible with Go's version guarantees.

#### Environment details

  - Programming language: Golang
  - OS: macOS
  - Language runtime version: latest

#### Steps to reproduce

  1. update google.golang.org/api from 0.219.0 to 0.220.0
 

#### Version Request

- Would you please reduce the minimum Go version just as https://github.com/grpc/grpc-go/pull/7831/?
- Would you please drop the toolchain directive, unless it's necessary?

Related/Relevant Issues
- https://github.com/golang/go/issues/65847 
- https://github.com/grpc/grpc-go/pull/7831/
- https://github.com/InfluxCommunity/influxdb3-go/issues/133

quartzmo:
Related to https://github.com/googleapis/google-cloud-go/issues/11468.

matthewhartstonge:
I've created a minimal repro: https://github.com/matthewhartstonge/exp-gomod-toolchain-resolution

But yeah, the issue does seem to be the `otel` upstream dep

quartzmo:
@matthewhartstonge Do you know why the minimum in this repo is `1.22.7`, while the otel minimum is `1.22.0`? If not, I'll try to investigate.

matthewhartstonge:
yarp, essentially what you found in #3011 due to `go: cloud.google.com/go/auth@v0.14.1 requires go@1.22.7, but 1.22 is requested`

quartzmo:
Note that opentelemetry-go reduced its version to `1.22.0` in Release v1.34.0/v0.56.0/v0.10.0 https://github.com/open-telemetry/opentelemetry-go/pull/6073

quartzmo:
Discussion of patch version in golang/go/issues/65847 issue about toolchain: https://github.com/golang/go/issues/65847#issuecomment-2291002573

quartzmo:
[...]

PR #3011, opened 2025-02-11
chore(all): restore go 1.22 in go.mod

quartzmo:
fixes: #3009


codyoss:
I don't think we can make this change because of our dependencies have a patch level specified in thier go.mod.

quartzmo:
In Kokoro, it fails the go.mod diff check in latest-version (below).

```
+ cd /root/go/src/google.golang.org/api/
+ try3 go mod download
+ eval 'go mod download'
++ go mod download
+ ./internal/kokoro/vet.sh
+ [[ cloud-devrel/client-libraries/go/google-api-go-client/presubmit/latest-version != *\l\a\t\e\s\t\-\v\e\r\s\i\o\n* ]]
+ go mod tidy
+ tee /dev/stderr
+ read
+ git diff go.mod
diff --git a/go.mod b/go.mod
index 3e19e7de8b..b48b684637 100644
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module google.golang.org/api
 
-go 1.22
+go 1.22.7
+
+toolchain go1.23.5
```

matthewhartstonge:
I've created a minimal repro: https://github.com/matthewhartstonge/exp-gomod-toolchain-resolution

But yeah, the issue does seem to be the `otel` upstream dep.

Setting `v1.22.0` should solve it, but may force downstream to set the patch. 
But adding a 0 to the `go` directive is better than having to manage the `toolchain` directive.

quartzmo:
@codyoss https://github.com/golang/go/issues/65847

codyoss:
@quartzmo That talks to the toolchain aspect, but the Go version part is likely a seperate and maybe even more contentious topic.

PR #3012, opened 2025-02-11
Create go.yml

Jguest7:
first run


quartzmo:
This looks like a probe for malware.

PR #3050, opened 2025-03-05
fix(transport/http): Add user agent to request headers

### LINKED ISSUE #3049: transport/http: User agent is ignored
In google.golang.org/api/transport/http.NewClient, execution eventually reaches this code: 

https://github.com/googleapis/google-api-go-client/blob/88e8ed9bd77df46f04da575ef89d94f7aa11f1e4/transport/http/dial.go#L101-L107

Note that the user agent isn't set, whereas it is in the other code path:

https://github.com/googleapis/google-api-go-client/blob/88e8ed9bd77df46f04da575ef89d94f7aa11f1e4/transport/http/dial.go#L164

enocom:
Fixes #3049

### DIFF
--- a/transport/http/dial.go
+++ b/transport/http/dial.go
@@ -105,6 +105,9 @@ func newClientNewAuth(ctx context.Context, base http.RoundTripper, ds *internal.
 	if ds.RequestReason != "" {
 		headers.Set("X-goog-request-reason", ds.RequestReason)
 	}
+	if ds.UserAgent != "" {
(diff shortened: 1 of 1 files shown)


enocom:
Because the `NewClient` function returns an `*http.Client` with the `http.RoundTripper` interface, there's no good way to test this change without a much larger update to the tests, e.g., configuring a `httptest.Server` and verifying the request.

codyoss (approved):

PR #3062, opened 2025-03-13
Add support for Flex Start node pool provisioning model

adamoldak:
feat: add support for Flex Start node pool provisioning model

### DIFF
--- a/container/v1/container-api.json
+++ b/container/v1/container-api.json
@@ -5714,6 +5714,10 @@
           "format": "google-duration",
           "type": "string"
         },
+        "flexStart": {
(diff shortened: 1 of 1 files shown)


quartzmo:
Hi @adamoldak,

The `*-api.json` files in this repository are regularly updated (overwritten) by a batch job consuming upstream discovery document sources. Example: #3059. A manual edit won't stick. Do you want to update the upstream discovery document instead?

quartzmo:
Here is the [discovery document for container:v1](https://container.googleapis.com/$discovery/rest?version=v1).

PR #3075, opened 2025-03-20
fix(googleapi): add JSON array support to CheckResponseWithBody

quartzmo:
fixes: googleapis/gax-go#389

```
R.......R......R.....R.......R......R......R.....R.......R.......R.......R
2025/03/20 15:59:31 googleapi: Error 429: Resource has been exhausted (e.g. check quota).
```

### DIFF
--- a/googleapi/googleapi.go
+++ b/googleapi/googleapi.go
@@ -184,11 +184,10 @@ func CheckResponseWithBody(res *http.Response, body []byte) error {
 // may be a JSON object or JSON array, or may be something else.
 func errorReplyFromBody(body []byte) (*errorReply, error) {
 	jerr := new(errorReply)
-	peekChar := body[0]
-	if peekChar == '[' {
+	if body[0] == '[' {
 		// Attempt JSON array
-		jsonArr := &[]*errorReply{jerr}
-		err := json.Unmarshal(body, jsonArr)
+		jsonArr := []*errorReply{jerr}
+		err := json.Unmarshal(body, &jsonArr)
 		return jerr, err
 	}
 	// Attempt JSON object
--- a/googleapi/googleapi.go
+++ b/googleapi/googleapi.go
@@ -163,8 +163,7 @@ func CheckResponseWithBody(res *http.Response, body []byte) error {
 		return nil
 	}
 
-	jerr := new(errorReply)
-	err := json.Unmarshal(body, jerr)
+	jerr, err := errorReplyFromBody(body)
 	if err == nil && jerr.Error != nil {
 		if jerr.Error.Code == 0 {
 			jerr.Error.Code = res.StatusCode
@@ -179,6 +178,22 @@ func CheckResponseWithBody(res *http.Response, body []byte) error {
 		Body:   string(body),
 		Header: res.Header,
 	}
+}
+
+// errorReplyFromBody attempts to get the error from body. The body
+// may be a JSON object or JSON array, or may be something else.
+func errorReplyFromBody(body []byte) (*errorReply, error) {
+	jerr := new(errorReply)
+	peekChar := body[0]
+	if peekChar == '[' {
+		// Attempt JSON array
+		jsonArr := &[]*errorReply{jerr}
+		err := json.Unmarshal(body, jsonArr)
+		return jerr, err
+	}
+	// Attempt JSON object
+	err := json.Unmarshal(body, jerr)
+	return jerr, err
 }
 
 // IsNotModified reports whether err is the result of the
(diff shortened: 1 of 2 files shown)


googleapi/googleapi.go
```diff
@@ -181,6 +180,22 @@ func CheckResponseWithBody(res *http.Response, body []byte) error {
 	}
 }
 
+// errorReplyFromBody attempts to get the error from body. The body
+// may be a JSON object or JSON array, or may be something else.
+func errorReplyFromBody(body []byte) (*errorReply, error) {
+	jerr := new(errorReply)
```
noahdietz, line 186: Probably still want something like this:
```suggestion
        if len(body) == 0 {
          return nil, nil
        }
	jerr := new(errorReply)
```

googleapi/googleapi.go
```diff
@@ -181,6 +180,22 @@ func CheckResponseWithBody(res *http.Response, body []byte) error {
...
+	peekChar := body[0]
+	if peekChar == '[' {
```
noahdietz, line 188: nit: Since we don't need the variable anywhere else, maybe just the following:
```suggestion
	if body[0] == '[' {
```

or `if peekChar := body[0]; peekChar == '[' {`

googleapi/googleapi.go
```diff
@@ -181,6 +180,22 @@ func CheckResponseWithBody(res *http.Response, body []byte) error {
...
+		// Attempt JSON array
+		jsonArr := &[]*errorReply{jerr}
+		err := json.Unmarshal(body, jsonArr)
```
noahdietz, line 191: Nit: feels cleaner to let the variable be the value, and let the usage use the reference, at least in the case of a slice - just my two cents
```suggestion
		jsonArr := []*errorReply{jerr}
		err := json.Unmarshal(body, &jsonArr)
```

googleapi/googleapi.go
quartzmo: Given the conditional above (`if err == nil && jerr.Error != nil {`), we need to return non-nil `err` if `errorReply` is nil. I think it's covered by `Unmarshal` if body is empty.

googleapi/googleapi.go
quartzmo: Good catch, copy/paste error. Fixed.

googleapi/googleapi.go
quartzmo: Fixed, thanks!

googleapi/googleapi_test.go
```diff
@@ -312,6 +312,32 @@ Details:
...
+	{
+		// Case: Error in JSON object in body.
+		&http.Response{
+			StatusCode: http.StatusTooManyRequests,
+		},
+		`{"error":{"code": 429,"message": "Resource has been exhausted (e.g. check quota).","status": "RESOURCE_EXHAUSTED"}}`,
+		&Error{
+			Code:    http.StatusTooManyRequests,
+			Message: "Resource has been exhausted (e.g. check quota).",
+			Body:    "{\"error\":{\"code\": 429,\"message\": \"Resource has been exhausted (e.g. check quota).\",\"status\": \"RESOURCE_EXHAUSTED\"}}",
```
noahdietz, line 324: Use backtick so that we don't have to escape these double quotes everywhere.
```suggestion
			Body:    `{"error":{"code": 429,"message": "Resource has been exhausted (e.g. check quota).","status": "RESOURCE_EXHAUSTED"}}`,
```

googleapi/googleapi_test.go
```diff
@@ -312,6 +312,32 @@ Details:
...
+	{
+		// Case: Streaming error in JSON array in body.
+		&http.Response{
+			StatusCode: http.StatusTooManyRequests,
+		},
+		`[{"error":{"code": 429,"message": "Resource has been exhausted (e.g. check quota).","status": "RESOURCE_EXHAUSTED"}}]`,
+		&Error{
+			Code:    http.StatusTooManyRequests,
+			Message: "Resource has been exhausted (e.g. check quota).",
+			Body:    "[{\"error\":{\"code\": 429,\"message\": \"Resource has been exhausted (e.g. check quota).\",\"status\": \"RESOURCE_EXHAUSTED\"}}]",
```
noahdietz, line 337: Same thing
```suggestion
			Body:    `[{"error":{"code": 429,"message": "Resource has been exhausted (e.g. check quota).","status": "RESOURCE_EXHAUSTED"}}]`,
```

noahdietz (approved):
Approved with two nits

googleapi/googleapi.go
quartzmo: Done, thanks!

googleapi/googleapi_test.go
quartzmo: Lol, it was easier to copy in the escaped quotes from console than to use backticks and remove them, so no labor lost here. But I see your point about consistency.

PR #3130, opened 2025-04-30
Add ability to specify []otelhttp.Opts for transport when create new HTTP client

surik:


quartzmo:
@surik Can you please open a [feature request issue](https://github.com/googleapis/google-api-go-client/issues/new/choose) in this repo explaining your need for [otelhttp.Option](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#Option) in the HTTP Client? Thank you.

transport/grpc/dial.go
```diff
@@ -400,7 +400,13 @@ func addOpenTelemetryStatsHandler(opts []grpc.DialOption, settings *internal.Dia
 	if settings.TelemetryDisabled {
 		return opts
 	}
-	return append(opts, grpc.WithStatsHandler(otelGRPCStatsHandler()))
+	otelOpts := []otelgrpc.Option{}
+	for _, opt := range settings.OpenTelemetryOpts {
+		if opt, ok := opt.(otelgrpc.Option); ok {
```
quartzmo, line 405: I think we should return an error when typecasting to both `otelgrpc.Option` and `otelhttp.Option` fails. That way users will understand immediately (at client init) that they have passed an invalid type. This can probably be done in `DialSettings.Validate`.  You can similarly handle the conversions in new accessors: `DialSettings.OtelGRPCOpts` and `DialSettings.OtelHTTPOpts`, in order to keep this logic out of `transport`.

surik: Done here https://github.com/googleapis/google-api-go-client/pull/3130/commits/899af8484d6587931824d127259c0bd9580ecfca

quartzmo (dismissed):

internal/settings.go
```diff
@@ -178,6 +185,18 @@ func (ds *DialSettings) Validate() error {
 	if ds.ImpersonationConfig != nil && len(ds.ImpersonationConfig.Scopes) == 0 && len(ds.Scopes) == 0 {
 		return errors.New("WithImpersonatedCredentials requires scopes being provided")
 	}
+	for _, opt := range ds.OpenTelemetryOpts {
+		switch o := opt.(type) {
+		case otelhttp.Option:
+			ds.OpenTelemetryOptsHTTP = append(ds.OpenTelemetryOptsHTTP, o)
+		case otelgrpc.Option:
+			ds.OpenTelemetryOptsGRPC = append(ds.OpenTelemetryOptsGRPC, o.(otelgrpc.Option))
```
quartzmo, line 193: internal/settings.go:193:64: type assertion to the same type: o already has type otelgrpc.Option (S1040)

transport/http/dial.go
```diff
@@ -306,7 +306,7 @@ func addOpenTelemetryTransport(trans http.RoundTripper, settings *internal.DialS
 	if settings.TelemetryDisabled {
 		return trans
 	}
-	return otelhttp.NewTransport(trans)
+	return otelhttp.NewTransport(trans, settings.OpenTelemetryOptsHTTP...)
```
codyoss, line 309: For both this and the gRPC package this change is fine to make, but it won't actually enable the feature for most users -- that is because these packages, by default, delegate to our new auth library which has its own transport packages: https://pkg.go.dev/cloud.google.com/go/auth

At the very least I think we should add some TODOs and keep the issue open until this is plumbed all the way through to the new auth packages. See the `newClientNewAuth` function in this file as an example of how we delegate calls to this new library.

surik: Thank you @codyoss! I added a todo.

quartzmo (approved):

PR #3166, opened 2025-05-22
fix(internaloption): AuthCreds should honor WithoutAuthentication

codyoss:
In this case the function will return a nil error and a nil credential. This API is currently only used by storage, whom have asked for this usage change. Also, fixed godoc that was not rendering properly before.

Internal Bug: 419426629


codyoss:
cc @BrennaEpp

BrennaEpp (dismissed):
Thank you!

internal/creds.go
```diff
@@ -47,6 +47,8 @@ func Creds(ctx context.Context, ds *DialSettings) (*google.Credentials, error) {
 // options provided via [option.ClientOption], including legacy oauth2/google
 // options. If there are no applicable options, then it returns the result of
 // [cloud.google.com/go/auth/credentials.DetectDefault].
+// Note: If NoAuth is true(when option.WithoutAuthentication is passed), then no
```
noahdietz, line 50: Nit: should this be linked like others in this commentary?
```suggestion
// Note: If NoAuth is true(when [option.WithoutAuthentication] is passed), then no
```

codyoss: done.

noahdietz (approved):

PR #3172, opened 2025-05-27
docs(internaloption): fix AuthCreds bulleted list format

noahdietz:
Currently broken, need two leading spaces before list item marker.

image

Switched to dash (aligns with gofmt rec and other examples in code).


noahdietz:
Fixed by #3166

PR #3258, opened 2025-08-04
fix(gensupport): fix transferChunk to return un-cancelled response which fixes race condition.

meet2mky:
The previous implementation used context.WithTimeout to manage timeouts for each chunk upload and cancelled the upload when the response was received. This introduced a race condition when the response was not yet processed by caller and context was cancelled.
This fix does the following:
 * Starts a timer and the upload request in a separate goroutine.
* Uses a select statement to wait for either the upload to complete, the timer to fire, or the context to be cancelled.
* Explicitly cancels the upload request if the timer fires first, preventing the race condition and ensuring resources are cleaned up correctly.
* Timer now watches for chunkRetryDeadline as well if the retries are not enabled to correctly cancel the chunkUpload after 32s.


internal/gensupport/resumable.go
```diff
@@ -127,34 +127,74 @@ func (rx *ResumableUpload) reportProgress(old, updated int64) {
...
-	var rCtx context.Context
-	var cancel context.CancelFunc
-
-	rCtx = ctx
+// It uses a goroutine to perform the upload and a timer to enforce ChunkTransferTimeout or ChunkRetryDeadline.
+func (rx *ResumableUpload) transferChunk(ctx context.Context, chunkRetryDeadline time.Duration, chunk io.Reader, off, size int64, done bool) (*http.Response, error) {
 	if rx.ChunkTransferTimeout != 0 {
-		rCtx, cancel = context.WithTimeout(ctx, rx.ChunkTransferTimeout)
-		defer cancel()
+		chunkRetryDeadline = rx.ChunkTransferTimeout
```
raj-prince, line 133: Why are overriding `retryDeadline` with `chunkTransferTimeout`? - I feel, we should not invoke stall logic (timer & go-routine) creation if ChunkTransferTimeout is not set.

meet2mky: Why shouldn't we? Earlier if it gets stalled beyond 32s then it would just stay stalled but now since we have the timer why not cancel it? Because Ideally, this should complete in 32s right?

internal/gensupport/resumable.go
```diff
@@ -127,34 +127,91 @@ func (rx *ResumableUpload) reportProgress(old, updated int64) {
...
+	go func() {
+		res, err := rx.doUploadRequest(rCtx, chunk, off, size, done)
+		resultCh <- uploadResult{res: res, err: err}
+	}()
+
+	// Wait for timer to fire or result channel to have the uploadResult or ctx to be cancelled.
+	select {
+	case <-ctx.Done():
+		// Context is cancelled for the overall upload.
+		cancel()
```
raj-prince, line 178: nit: just check, if it's safe to call cancel() over the cancelled context.

raj-prince (approved):
LGTM overall, few nits.

internal/gensupport/resumable.go
```diff
@@ -127,34 +127,91 @@ func (rx *ResumableUpload) reportProgress(old, updated int64) {
...
+		// drain the resultCh to ensure go routine is finished.
```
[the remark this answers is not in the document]
meet2mky: Done

internal/gensupport/resumable.go
```diff
@@ -127,34 +127,72 @@ func (rx *ResumableUpload) reportProgress(old, updated int64) {
...
+		// Context is cancelled for the overall upload.
+		cancel()
+		// Drain the resultCh to ensure go routine is finished.
+		<-resultCh
+		return nil, ctx.Err()
+	case <-timer.C:
+		// timer fired first so we cancel the chunk upload and return
+		// deadline exceeded for this chunk upload.
+		cancel()
+		// Drain the resultCh to ensure go routine is finished.
```
tritone, line 178: Can we also note that calling cancel will force the HTTP request to return, ensuring that there is something to drain? I had to look at the code to convince myself this was true :)

internal/gensupport/resumable.go
```diff
@@ -127,34 +127,72 @@ func (rx *ResumableUpload) reportProgress(old, updated int64) {
...
+		<-resultCh
+		return nil, context.DeadlineExceeded
+
+	case result := <-resultCh:
+		// The upload finished before the timer fired. Stop the timer to prevent a resource leak.
+		if !timer.Stop() {
```
tritone, line 184: IIUC this is no longer necessary in supported go versions. See https://go.dev/wiki/Go123Timer

codyoss (approved):

PR #3317, opened 2025-09-24
chore(google-api-go-generator): add aiplatform:v1beta1 to skipAPIGeneration

quartzmo:
Restore aiplatform:v1beta1 after b/446965637 is resolved.


google-api-go-generator/gen.go
```diff
@@ -79,6 +79,8 @@ var skipAPIGeneration = map[string]bool{
 	"integrations:v1":      true,
 	"sql:v1beta4":          true,
 	"datalineage:v1":       true,
+	// Restore aiplatform:v1beta1 after b/446965637 is resolved.
```
codyoss, line 82: nit: can you file a bug and link it here to make sure we get back to this.

PR #3356, opened 2025-10-28
feat(option): Deprecate unsafe credentials JSON loading options

quartzmo:
Add safer credentials JSON loading options.

Add option.WithAuthCredentialsFile and option.WithAuthCredentialsJSON
to mitigate a security vulnerability where credential configurations
from untrusted sources could be used without validation. These new
functions require the credential type to be explicitly specified.

Deprecate the less safe option.WithCredentialsFile and
 option.WithCredentialsJSON functions.

refs: googleapis/google-cloud-go#13397

### DIFF
--- a/aiplatform/v1/aiplatform-api.json
+++ b/aiplatform/v1/aiplatform-api.json
@@ -19456,7 +19456,7 @@
                       "type": "string"
                     },
                     "updateMask": {
-                      "description": "Required. The update mask applies to the resource. For the `FieldMask` definition, see google.protobuf.FieldMask. Input format: `{paths: \"${updated_filed}\"}` Updatable fields: * `encryption_spec.kms_key_name`",
+                      "description": "Required. The update mask applies to the resource. For the `FieldMask` definition, see google.protobuf.FieldMask. Input format: `{paths: \"${updated_field}\"}` Updatable fields: * `encryption_spec.kms_key_name` * `display_name` * `software_config.post_startup_script_config.post_startup_script` * `software_config.post_startup_script_config.post_startup_script_url` * `software_config.post_startup_script_config.post_startup_script_behavior` * `software_config.env` * `software_config.colab_image.release_name`",
                       "format": "google-fieldmask",
                       "location": "query",
                       "type": "string"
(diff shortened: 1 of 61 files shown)


shollyman (approved):
Should there be tests in this PR where the file/json contents and the credential type are mismatched?  I suspect they belong closer to wherever the validation logic lives, but I wanted to doublecheck with you.

option/option.go
```diff
@@ -18,6 +18,51 @@ import (
...
+	//
+	// IMPORTANT:
+	// This credential type does not validate the credential configuration. A security
+	// risk occurs when a credential configuration configured with malicious urls
+	// is used.
+	// You should validate credential configurations provided by untrusted sources.
+	// See [Security requirements when using credential configurations from an external
+	// source] https://cloud.google.com/docs/authentication/external/externally-sourced-credentials
+	// for more details.
+	Unknown CredentialsType = internal.Unknown
```
sai-sunder-s: same here. can we avoid adding unknown?

quartzmo: Removed, thanks!

idtoken/idtoken.go
```diff
@@ -141,12 +190,12 @@ func newTokenSource(ctx context.Context, audience string, ds *internal.DialSetti
...
 
 func tokenSourceFromBytes(ctx context.Context, data []byte, audience string, ds *internal.DialSettings) (oauth2.TokenSource, error) {
-	allowedType, err := getAllowedType(data)
+	credType, err := credentialstype.GetCredType(data)
 	if err != nil {
 		return nil, err
 	}
-	switch allowedType {
-	case serviceAccount:
+	switch credType {
```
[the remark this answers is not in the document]
quartzmo: This limited set of types preserves [existing behavior](https://github.com/googleapis/google-api-go-client/blob/main/idtoken/idtoken.go#L148-L192). Also, this `idtoken` package is legacy, and users should migrate to [cloud.google.com/go/auth/idtoken](https://pkg.go.dev/cloud.google.com/go/auth/idtoken).

sai-sunder-s (approved):

PR #3373, opened 2025-11-13
feat(transport/grpc): add a getter for grpcconnpool size

sushanb:
The Bigtable Go client library allows users to configure the underlying gRPC connection pool size using option.WithGRPCConnectionPool(size) when creating a new client. However, the Bigtable client code itself has no direct way to access the size value that the user provided. This is because the option is applied to an internal DialSettings struct within the google.golang.org/api dependencies, and the final pool size is not exposed.

We need to access this size to use the custom Bigtable Channel Pool.

### DIFF
--- a/transport/grpc/dial.go
+++ b/transport/grpc/dial.go
@@ -552,6 +552,16 @@ func processAndValidateOpts(opts []option.ClientOption) (*internal.DialSettings,
 	return &o, nil
 }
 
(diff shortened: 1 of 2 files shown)


sushanb:
@nimf FYI if this is the right way to fix this.

mutianf (approved):

shollyman:
Closing this as it was superseded by https://pkg.go.dev/google.golang.org/api/option/internaloption#UnsafeResolver

PR #3420, opened 2025-12-23
fix(option): remove option.WithAuthCredentials from validation

quartzmo:
Do not raise the multiple credential options provided error when
option.WithAuthCredentials is provided with other auth configuration
options, as it is a special case used internally for oauth2 type converversions.

fixes: googleapis/google-cloud-go#13503


internal/settings.go
shollyman: Generally fine with leaving this aspect of ds.AuthCredentials intact.  There's still some auth scenarios that are indirect (e.g. pass in an authenticated HTTP client), so we can't validate all scenarios.

PR #3491, opened 2026-02-11
feat(transport/grpc): add function to get reason for direct path incompatibility

cpriti-os:
Adds DirectPathStatus and CheckWithReason to categorize why DirectPath is used or disabled  into fallback reasons. This enables better observability.

### DIFF
--- a/transport/grpc/dial_test.go
+++ b/transport/grpc/dial_test.go
@@ -415,6 +415,13 @@ func TestCheckAuthStatus(t *testing.T) {
 			},
 			want: statusTokenFetchError,
 		},
+		{
+			name: "Nil Token - Incompatible",
+			ds: &internal.DialSettings{
+				TokenSource: &mockTokenSource{token: nil}, // Token() returns (nil, nil)
+			},
+			want: statusNilToken,
+		},
 	}
 
 	for _, tt := range tests {
@@ -546,6 +553,17 @@ func TestCheckDirectPathStatus(t *testing.T) {
 			},
 			onGCE: true,
 			want:  statusTokenFetchError,
+		},
+		{
+			name: "Nil token returned",
+			opts: []option.ClientOption{
+				internaloption.EnableDirectPath(true),
+				internaloption.EnableDirectPathXds(),
+				validEndpoint,
+				option.WithTokenSource(&mockTokenSource{token: nil}),
+			},
+			onGCE: true,
+			want:  statusNilToken,
 		},
 		{
 			name: "Incompatible credentials - Not compute metadata",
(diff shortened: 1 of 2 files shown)


cpriti-os:
> @cpriti-os Why is this component here in https://github.com/googleapis/google-api-go-client/tree/main/transport/grpc and not for example in the newer equivalent package https://github.com/googleapis/google-cloud-go/tree/main/auth/grpctransport ?

Thanks for pointing out, couldn't find it when I raised this, will try to move the code there.

transport/grpc/dial_test.go
```diff
@@ -258,3 +264,336 @@ func TestGRPCAPIKey_GetRequestMetadata(t *testing.T) {
...
+
+			ds := &internal.DialSettings{EnableDirectPathXds: tt.opt}
+			if got := isDirectPathXdsUsed(ds); got != tt.want {
+				t.Errorf("isDirectPathXdsUsed() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestIsDirectPathEnabled(t *testing.T) {
```
krishnamd-jkp: Are these tests relevant to the changes?

transport/grpc/dial_test.go
```diff
@@ -258,3 +264,336 @@ func TestGRPCAPIKey_GetRequestMetadata(t *testing.T) {
...
+				t.Errorf("isDirectPathEnabled() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+func TestAuthReason(t *testing.T) {
+	ctx := context.Background()
+	detector := NewDirectPathStatus()
+
+	// A token that satisfies the GCE Metadata requirement
```
krishnamd-jkp: nit: missing period. Other places as well

shollyman (commented):
I'm still a bit confused a bit by how this code is intended to be leveraged, given the mix of resolving dial options, checking resolved grpc, and resolving auth configuration.  Is there a doc that can help me understand this a bit better?  We've had a few similar requests recently around complexities resolving client options, and I'm trying to figure out if there's common functionality we should be building out.  This very detailed exported function that appears to rationalize why a behavior didn't happen, and I wonder if we can simplify it with some underlying designs.

cpriti-os:
> I'm still a bit confused a bit by how this code is intended to be leveraged, given the mix of resolving dial options, checking resolved grpc, and resolving auth configuration. Is there a doc that can help me understand this a bit better? We've had a few similar requests recently around complexities resolving client options, and I'm trying to figure out if there's common functionality we should be building out. This very detailed exported function that appears to rationalize why a behavior didn't happen, and I wonder if we can simplify it with some underlying designs.

@shollyman thank you for checking this. Our current goal is pretty simple, provide as granular a reason as possible as to why the client chose not to use Direct Path. The added function would be called by GCS Go SDK while creating a new gRPC client and pass the received reason to the backend using headers. The backend can then log this telemetry and  in the near future reject a cloudpath request with the reason (in this PR) to help customer resolve the issue more efficiently. I have also provided a one pager discussing this in the chat space. Let me know if this answers the question. Happy to discuss more.

PR #3667, opened 2026-07-22
fix(transport): allow configuring HTTP/2 ReadIdleTimeout

### LINKED ISSUE #3510: configureHTTP2 methods default read idle timeout is an absolute pain when dealing with androidpublisher service.
#### Environment details

  - Programming language: golang
  - OS: linux
  - Language runtime version: 1.25
  - Package version: v0.267.0

#### Steps to reproduce
```
		gcpcreds := option.WithCredentialsJSON(envx.Base64([]byte("{}"), _eg.EnvUnsafeGcloudADCB64))
		scopes := option.WithScopes(androidpublisher.AndroidpublisherScope)


		client, endpoint, err := htransport.NewClient(ctx, gcpcreds, dialer, keepalive, scopes)
		if err != nil {
			return err
		}

		service, err := androidpublisher.NewService(ctx, option.WithHTTPClient(client), option.WithEndpoint(endpoint))
		if err != nil {
			return fmt.Errorf("play store init failed: %w", err)
		}
```
configureHTTP2 hardcodes the timeout to 31 seconds and the client which htransport.NewClient returns is impossible to configure. as a result one has to entirely configure a http client manually, which is awful.

SWITCHin2:
Fixes: #3510

Add option.WithHTTP2ReadIdleTimeout so callers can override the timeout while still using the library's credential, endpoint and mTLS setup. The default of 31s is preserved when the option is unset or non-positive.


SWITCHin2:
Hey @shollyman Can you please help me with requesting an review for the PR.

PR #3668, opened 2026-07-23
fix(idtoken): validate ES256 signature length before slicing

prasanna8585:
## Problem

`idtoken.Validator.Validate()` panics with an unrecovered runtime error when given a JWT signed with `alg: ES256` whose signature segment decodes to fewer than 32 bytes. Any service using this package to validate incoming Google-issued ID tokens (e.g. Cloud Run, Cloud Functions, custom services authenticating callers via bearer tokens) can be crashed by a single malformed ~120-byte JWT.

## Root Cause

`idtoken/validate.go`, `validateES256`:
```
r := big.NewInt(0).SetBytes(sig[:es256KeySize])   // panics if len(sig) < 32
s := big.NewInt(0).SetBytes(sig[es256KeySize:])
```

`findMatchingKey` only checks that the JWT's `kid` header matches a known key ID — it does not validate the signature at that point. The JWKS endpoint (`googleIAPCertsURL`) is public and unauthenticated by design, so any attacker can retrieve a real, currently valid `kid` with a single unauthenticated GET request, then craft a JWT with that `kid` and a signature under 32 bytes to trigger the panic.

## Fix

Adds an exact-length check before slicing:
```
if len(sig) != es256KeySize*2 {
    return fmt.Errorf("idtoken: invalid ES256 signature length: got %d bytes, want %d", len(sig), es256KeySize*2)
}
```

This also incidentally fixes a related correctness issue: signatures longer than 64 bytes were previously silently truncated to the first 64 bytes without error (only the exact-length check catches both directions).

## Testing

Two new regression tests in `idtoken/validate_test.go`:
- `TestValidateES256_ShortSignature`: a 3-byte signature must return an error, not panic
- `TestValidateES256_OverlongSignature`: a 65-byte signature must also return an error

Both use the package's real, unmodified, exported `Validator.Validate()` entry point via the existing `RoundTripFn`/`commonToken` test helpers — no reimplementation.

---

### DIFF
--- a/idtoken/validate.go
+++ b/idtoken/validate.go
@@ -229,6 +229,14 @@ func (v *Validator) validateES256(ctx context.Context, keyID string, hashedConte
 		X:     new(big.Int).SetBytes(dx),
 		Y:     new(big.Int).SetBytes(dy),
 	}
+	// Validate the signature length before slicing. A JWT with a signature
+	// segment that decodes to fewer than es256KeySize*2 bytes would otherwise
+	// cause an unrecovered panic ("slice bounds out of range") here, since
+	// sig[:es256KeySize] and sig[es256KeySize:] assume exactly two
+	// es256KeySize-byte components (r and s) with no length check.
+	if len(sig) != es256KeySize*2 {
+		return fmt.Errorf("idtoken: invalid ES256 signature length: got %d bytes, want %d", len(sig), es256KeySize*2)
+	}
 	r := big.NewInt(0).SetBytes(sig[:es256KeySize])
 	s := big.NewInt(0).SetBytes(sig[es256KeySize:])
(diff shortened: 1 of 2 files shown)


prasanna8585:
Hi Team,

Just checking in to see if you've had a chance to look at the patch strategy for this. I'm happy to spin up the code changes for the fix in a private fork if that helps speed things up.

Thanks again for your time!

prasanna8585:
Hi Team,

Just checking in to see if you've had a chance to look at the patch strategy for this. I'm happy to spin up the code changes for the fix in a private fork if that helps speed things up.

prasanna8585:
Hi Team,

Just checking in to see if you've had a chance to look at the patch strategy for this.

prasanna8585:
Hi Team,

Just checking in to see if you've had a chance to look at the patch strategy for this.

